PR #20, opened 2020-10-27
Remove community.general as a test requirement

Andersson007:
##### SUMMARY
Remove community.general as a test requirement


felixfontein:
In that case you also have to remove it from tests/utils/shippable/shippable.sh. (This file here is informational at the moment, it's not used by ansible-test so far.)

I don't think you can really drop this requirement though, except if you want to drop the OpenSUSE CI targets.

Andersson007:
@felixfontein thanks for the information!

PR #27, opened 2020-11-17
Update README

Andersson007:
##### SUMMARY
To prevent cases when users put it in requirements and forget about it.

##### ISSUE TYPE
- Docs Pull Request

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,6 @@ You can include it in a `requirements.yml` file and install it via `ansible-gala
 ---
 collections:
   - name: community.postgresql
-    version: 1.0.0
 ```
 
 You can also download the tarball from Ansible Galaxy and install the collection manually wherever you need.
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ You can include it in a `requirements.yml` file and install it via `ansible-gala
 ---
 collections:
   - name: community.postgresql
-    version: 0.1.0
+    version: 1.0.0
 ```
 
 You can also download the tarball from Ansible Galaxy and install the collection manually wherever you need.
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -1,6 +1,6 @@
 namespace: community
 name: postgresql
-version: 1.0.1
+version: 1.1.0
 readme: README.md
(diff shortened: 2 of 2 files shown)


galaxy.yml
```diff
@@ -1,6 +1,6 @@
 namespace: community
 name: postgresql
-version: 1.0.0
+version: 1.0.1
```
felixfontein, line 3: Once ansible-test has proper version_added tests (like ansible/ansible has), it's better to bump it to the next minor release 1.1.0. If you then decide that a patch release is needed reduce the version number before releasing it (to 1.0.1), release it, and re-bump to 1.1.0.

But as long as it's a WIP PR, no need to worry :)

README.md
```diff
@@ -58,7 +58,7 @@ You can include it in a `requirements.yml` file and install it via `ansible-gala
 ---
 collections:
   - name: community.postgresql
-    version: 0.1.0
+    version: 1.0.0
```
felixfontein, line 61: I think it's better to leave `version:` off, otherwise people will never get updates because they won't update the version.

Andersson007: good suggestion! I copy-pasted this from community.mysql. So, i'll remove this from there as well, thanks!

galaxy.yml
Andersson007: according to https://github.com/ansible/community/wiki/ReleasingCollections, we should adjust this to the next expected version. So, should we adjust the manual correspondingly?

felixfontein: It depends on whether you want to use the version_added tests in the future, or not. If you want to use them, the tests somehow need to know which version they need to ensure in version_added. For that, you need the next minor/major release in galaxy.yml. But if you don't want to use that test (or only change it once that test actually is available), no need to change now.

Andersson007: Ah, ok, thanks for the explaining, now got it. anyway, better to clarify it in the release guide

Andersson007: i'll leave it as it is now, then will see

Andersson007: put the initial back

Issue #31, opened 2020-12-16
Unable to dynamically use/substitute positional_args to NULLs in postgresql_query


Iridescens:
##### SUMMARY

Unable to dynamically use/substitute positional_args to NULLs in postgresql_query

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

community.general.postgresql_query

##### ANSIBLE VERSION

```paste below
ansible 2.9.14
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/var/lib/awx/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.6/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.6.8 (default, Apr 16 2020, 01:36:27) [GCC 8.3.1 20191121 (Red Hat 8.3.1-5)]
```

##### CONFIGURATION

```paste below
DEFAULT_STDOUT_CALLBACK(/etc/ansible/ansible.cfg) = yaml
HOST_KEY_CHECKING(/etc/ansible/ansible.cfg) = False
```

##### OS / ENVIRONMENT

postgresql11-server-11.9-1PGDG.rhel7.x86_64
NAME="Red Hat Enterprise Linux Server"
VERSION="7.7 (Maipo)"

##### STEPS TO REPRODUCE

Try to dynamically upsert some NULLs into tables

```yaml
community.general.postgresql_query:
        login_host: "{{ awx_request_audit_db_host }}"
        login_user: "{{ awx_request_audit_db_user }}"
        login_password: "{{ awx_request_audit_db_pass }}"
        db: "{{ awx_request_audit_db_name }}"
        query: >
          INSERT INTO request_id_table(request_id)
           VALUES (%s) ON CONFLICT ON CONSTRAINT request_id_table_pk DO NOTHING;
          INSERT INTO requests(request_id,device_hostname,request_type,error_step,config_snapshot,awx_user_name)
           VALUES (%s,%s,%s,%s,%s,%s) ON CONFLICT ON CONSTRAINT requests_pk DO UPDATE
           SET error_step=NULL, config_snapshot=EXCLUDED.config_snapshot, awx_user_name=EXCLUDED.awx_user_name;
          INSERT INTO add_bd_table(request_id,bd_id,bd_vni_id,bd_routedist,bd_subinterface,bd_vlan_id,bd_if_encap)
[...]

PR #32, opened 2020-12-16
postgresql_query: add explanation how to pass undefined variables as NULLs to positional_args

Andersson007:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.postgresql/issues/31

postgresql_query: add explanation how to pass undefined variables as NULLs to positional_args

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
postgresql_query

### DIFF
--- a/plugins/modules/postgresql_query.py
+++ b/plugins/modules/postgresql_query.py
@@ -27,7 +27,6 @@ options:
     description:
     - List of values to be passed as positional arguments to the query.
       When the value is a list, it will be converted to PostgreSQL array.
-    - To pass NULL value, use C(null).
     - Mutually exclusive with I(named_args).
     type: list
     elements: raw
@@ -35,7 +34,6 @@ options:
     description:
     - Dictionary of key-value arguments to pass to the query.
       When the value is a list, it will be converted to PostgreSQL array.
-    - To pass NULL value, use C(null).
     - Mutually exclusive with I(positional_args).
     type: dict
   path_to_script:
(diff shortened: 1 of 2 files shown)


plugins/modules/postgresql_query.py
```diff
@@ -172,6 +172,24 @@
...
+# If you use a variable in positional_args / named_args that can
+# be undefined and you wish to set it as NULL, the constructions like
+# "{{ my_var if (my_var is defined) else none | default(none) }}"
+# will not work as expected substituting an empty string instead of NULL.
+# You should precheck such a value and define it as NULL when undefined.
+# For example:
+- name: When undefined, set to NULL
+  set_fact:
+    my_var: NULL
+    when: my_var is undefined
```
felixfontein, line 185:
```suggestion
  when: my_var is undefined
```

Andersson007: thanks @felixfontein

plugins/modules/postgresql_query.py
```diff
@@ -172,6 +172,24 @@
     search_path:
     - app1
     - public
+
+# If you use a variable in positional_args / named_args that can
+# be undefined and you wish to set it as NULL, the constructions like
+# "{{ my_var if (my_var is defined) else none | default(none) }}"
+# will not work as expected substituting an empty string instead of NULL.
```
felixfontein, line 179:
```suggestion
# will not work as expected substituting an empty string instead of NULL.
# If possible, we suggest to use Ansible's DEFAULT_JINJA2_NATIVE configuration
# (https://docs.ansible.com/ansible/latest/reference_appendices/config.html#default-jinja2-native).
# Enabling it fixes this problem. If you cannot enable it, the following workaround
# can be used.
```

felixfontein: (https://github.com/ansible/ansible/issues/73010#issuecomment-747445941)

Andersson007:
@felixfontein thanks for reviewing!

Issue #48, opened 2021-01-29
postgresql_set in check mode fails on non-numeric values containing `B`


k3rni:
##### SUMMARY
When running in check mode, `postgresql_set` throws an exception if the value to set contained the letter `B`, or other binary units like `KB, MB`, but was **not** numeric otherwise.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
postgresql_set

##### ANSIBLE VERSION
```
ansible 2.10.5
  config file = /home/k3rni/Source/ansible-bug-b/ansible.cfg
  configured module search path = ['/home/k3rni/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/k3rni/.asdf/installs/python/3.8.6/lib/python3.8/site-packages/ansible
  executable location = /home/k3rni/.asdf/installs/python/3.8.6/bin/ansible
  python version = 3.8.6 (default, Dec 17 2020, 13:42:12) [GCC 10.2.0]
```

##### CONFIGURATION
```paste below
DEFAULT_STDOUT_CALLBACK(/home/k3rni/Source/ansible-bug-b/ansible.cfg) = debug
```

##### OS / ENVIRONMENT
N/A

##### STEPS TO REPRODUCE
Minimal playbook, assumes postgres is on localhost. Provide `login_host` etc. otherwise.
```yaml
- hosts: localhost
  tasks:
    - community.general.postgresql_set:
        name: maintenance_work_mem
        value: 100MB

    - community.general.postgresql_set:
        name: archive_command
        value: '/usr/bin/touch %f' # NOTE: contains lowercase B, and not a very useful archive_command to set
```

Run in check mode:
```shell
ansible-playbook -i example.yml -C
```

##### EXPECTED RESULTS
Command completes successfully, reporting if the parameters need changing.

##### ACTUAL RESULTS

```
[WARNING]: No inventory was parsed, only implicit localhost is available
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'

PLAY [localhost] ****************************************************************************************************************************************************************

TASK [Gathering Facts] **********************************************************************************************************************************************************
ok: [localhost]

TASK [community.general.postgresql_set] *****************************************************************************************************************************************
ok: [localhost]

TASK [community.general.postgresql_set] *****************************************************************************************************************************************
[...]

PR #52, opened 2021-02-01
postgresql_set: fix crashes caused by values containing 'b'

Andersson007:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.postgresql/issues/48

I also made additional tests tests/integration/targets/postgresql_set/tasks/options_coverage.yml

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
postgresql_set

### DIFF
--- a/plugins/modules/postgresql_set.py
+++ b/plugins/modules/postgresql_set.py
@@ -240,28 +240,51 @@ def pretty_to_bytes(pretty_val):
     # if the value contains 'B', 'kB', 'MB', 'GB', 'TB'.
     # Otherwise it returns the passed argument.
 
+    # It's sometimes possible to have an empty values
+    if not pretty_val:
+        return pretty_val
+
+    # If the first char is not a digit, it does not make sense
+    # to parse further, so just return a passed value
+    if not pretty_val[0].isdigit():
+        return pretty_val
+
+    # If the last char is not an alphabetical symbol, it means that
+    # it does not contain any suffices, so no sense to parse further
+    if not pretty_val[-1].isalpha():
+        return pretty_val
+
+    # Extract digits
+    num_part = []
+    for c in pretty_val:
+        # When we reach the first non-digit element,
+        # e.g. in 1024kB, stop iterating
+        if not c.isdigit():
+            break
+
+        if c.isdigit():
+            num_part.append(c)
+
+    num_part = ''.join(num_part)
+
     val_in_bytes = None
 
     if len(pretty_val) >= 2:
         if 'kB' in pretty_val[-2:]:
-            num_part = int(''.join(d for d in pretty_val if d.isdigit()))
             val_in_bytes = num_part * 1024
 
-        elif 'MB' in pretty_val.upper()[-2:]:
-            num_part = int(''.join(d for d in pretty_val if d.isdigit()))
+        elif 'MB' in pretty_val[-2:]:
             val_in_bytes = num_part * 1024 * 1024
 
-        elif 'GB' in pretty_val.upper()[-2:]:
-            num_part = int(''.join(d for d in pretty_val if d.isdigit()))
+        elif 'GB' in pretty_val[-2:]:
             val_in_bytes = num_part * 1024 * 1024 * 1024
 
-        elif 'TB' in pretty_val.upper()[-2:]:
-            num_part = int(''.join(d for d in pretty_val if d.isdigit()))
+        elif 'TB' in pretty_val[-2:]:
             val_in_bytes = num_part * 1024 * 1024 * 1024 * 1024
 
-        elif len(pretty_val) >= 1 and 'B' in pretty_val.upper()[-1]:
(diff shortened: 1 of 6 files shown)


Andersson007:
recheck

Andersson007:
ready_for_review

Andersson007:
ready_for_review

plugins/modules/postgresql_set.py
```diff
@@ -240,33 +240,57 @@ def pretty_to_bytes(pretty_val):
...
-        val_in_bytes = num_part * 1024
+    if len(pretty_val) >= 2:
+        if 'kB' in pretty_val[-2:]:
+            val_in_bytes = num_part * 1024
 
-    elif 'MB' in pretty_val.upper():
-        num_part = int(''.join(d for d in pretty_val if d.isdigit()))
-        val_in_bytes = num_part * 1024 * 1024
+        elif 'MB' in pretty_val[-2:]:
+            val_in_bytes = num_part * 1024 * 1024
```
tcraxs, line 277: What about `mb`? Is this still covered like in line 249 with `pretty_val.upper()`?
Same is for `GB`, `TB` and `B`, please check for case sensitive.

Andersson007: @tcraxs yes, it should be covered, i consider this `.upper()` is extra because of line 345 in `main()`:
```
        for unit in POSSIBLE_SIZE_UNITS:
            if (value[:-2].isdigit() and unit in value[-2:]) or ('b' in value[-1] and value[:-1].isdigit()):
                value = value.upper()
```
and the function is called after this transformation so removed.
Also the CI tests cover the case you mentioned

tcraxs: Okay, perfect. I just oversaw this (also in the test cases).

tcraxs (approved):
LGTM

plugins/modules/postgresql_set.py
Andersson007: np, I also oversaw this when writing the module;)

plugins/modules/postgresql_set.py
```diff
@@ -182,7 +182,7 @@
 PG_REQ_VER = 90400
 
 # To allow to set value like 1mb instead of 1MB, etc:
-POSSIBLE_SIZE_UNITS = ("mb", "gb", "tb", "b")
+POSSIBLE_SIZE_UNITS = ("mb", "gb", "tb")
```
tcraxs: Is this intended? Why did you remove the `b` in this commit?

plugins/modules/postgresql_set.py
```diff
@@ -240,33 +240,57 @@ def pretty_to_bytes(pretty_val):
...
+    # Extract digits
+    num_part = []
+    for c in pretty_val:
+        # When we reach the first non-digit element,
+        # e.g. in 1024kB, stop iterating
+        if not c.isdigit():
+            break
+
+        if c.isdigit():
+            num_part.append(c)
```
k3rni, line 266: conditions in lines 262 and 265 are inverses of each other. So why not just use an `else`?

plugins/modules/postgresql_set.py
```diff
@@ -318,7 +342,7 @@ def main():
     # Allow to pass values like 1mb instead of 1MB, etc:
     if value:
         for unit in POSSIBLE_SIZE_UNITS:
-            if value[:-2].isdigit() and unit in value[-2:]:
+            if (value[:-2].isdigit() and unit in value[-2:]) or ('b' in value[-1] and value[:-1].isdigit()):
```
k3rni, line 345: this checks for the single `b` each time. shouldn't that check be outside of the for loop?

Andersson007: @k3rni yes, i'm in the business trip now, I'll rewrite the whole loop soon, I noticed that it's extra. Will fix it, thanks!

plugins/modules/postgresql_set.py
Andersson007: @tcraxs sorry for late response, I'm in a business trip, I added it in this PR but it's wrong (I'll add an explanation to the comment to this tuple), I'll continue to work on the PR soon, will let you know.

plugins/modules/postgresql_set.py
Andersson007: good spot! changed, thanks:)

plugins/modules/postgresql_set.py
Andersson007: Rewrote. I could've written one long condition but split it intentionally for better readability

plugins/modules/postgresql_set.py
Andersson007: I replaced the redundant loop where it's used, could you please take a look

Andersson007:
ready_for_review

tcraxs:
LGTM

Andersson007:
@tcraxs @k3rni thanks for reviewing!

PR #82, opened 2021-04-10
Fix fail_on_role check

gsauthof:
##### SUMMARY

This fixes the issue where the check is never reached because
the module is unconditionally exited, before, if the specified
role doesn't exist.

Which was effectively as if `fail_on_role` is always set to 'no'
plus the silencing of the warning. Which, of course, is quite a trap
and conflicts with the documentation, as well.

For symmetry and better troubleshooting that result now also includes
the executed queries (like the other call site).

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

postgresql_privs

##### ADDITIONAL INFORMATION

How to reproduce the original issue: try to grant some privilege to a non-existing role.

```paste below

```

### DIFF
--- a/changelogs/fragments/82-fix-fail-on-role.yml
+++ b/changelogs/fragments/82-fix-fail-on-role.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - postgresql_privs - fix ``fail_on_role`` check (https://github.com/ansible-collections/community.postgresql/pull/82).
--- a/plugins/modules/postgresql_privs.py
+++ b/plugins/modules/postgresql_privs.py
@@ -1121,13 +1121,11 @@ def main():
             roles = p.roles.split(',')
 
             if len(roles) == 1 and not role_exists(module, conn.cursor, roles[0]):
-                module.exit_json(changed=False)
-
                 if fail_on_role:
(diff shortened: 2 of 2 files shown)


Andersson007:
@gsauthof thanks for the PR!
could you please
1) add a changelog framgment https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment
2) would be also good to have the case covered with integration tests, see the content of `tests/integration/targets/postgresql_privs/tasks`

If any questions, feel free to ask

Andersson007:
(fyi: I close-reopened the PR to re-run the tests)

Andersson007:
Looks like an issue with Azure pipelines. I'll try to re-launch them later

gsauthof:
I've added the fragment.

I agree, it would be nice to include an integration test case. But currently I'm a bit time constrained. Thus, perhaps this would be something for another pull request.

Andersson007:
CC @tcraxs @andytom @kostiantyn-nemchenko @sebasmannem @ilicmilan @popov83

tcraxs (approved):
LGTM

Andersson007:
@gsauthof thanks for the fix (feel free to add CI tests), @tcraxs thanks for reviewing! (btw I sent you an email yesterday)

PR #119, opened 2021-07-26
Need a force (bool) option for postgresql_db. Otherwise fails to drop a db with state=absent only

asifiqbal:
##### SUMMARY

Fixes https://github.com/ansible-collections/community.postgresql/issues/109

Added a `force` (boolean) option with `state=absent`. Need to drop active connections first to be able to drop a database.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

postgresql_db needs a force option with state=absent or it will fail to drop a database with active connections

##### ADDITIONAL INFORMATION

1. When database does not exists, do nothing.
2. When database exists and has active connections, and force is True, it will drop the database
3. When database exists and has active connections, and force is False, it will fail to drop the database
4. When database exists and has no active connections, and force is False, it will drop the database
5. When database exists and has no active connections, and force is True, it will drop the database

```yaml
postgresql_db:
    name: mydb
    state: absent
    force: yes
```
`force:` is the new feature request. Without it a DB will not be removed with active connections
`force:` _default_ value is **False** and added to the documentation

### DIFF
--- a/changelogs/fragments/drop_db_with_force.yml
+++ b/changelogs/fragments/drop_db_with_force.yml
@@ -1,0 +1,3 @@
+postgresql_db name=mydb state=absent fails to remove mydb database if it has active connections.
+
+Added a force option (boolean) to remove active connections first and then remove the database.
--- a/plugins/modules/postgresql_db.py
+++ b/plugins/modules/postgresql_db.py
@@ -320,7 +320,7 @@ def db_exists(cursor, db):
 def db_dropconns(cursor, db):
     if cursor.connection.server_version >= 90200:
         """ Drop DB connections in Postgres 9.2 and above """
-        query_terminate = ("SELECT pg_terminate_backend(pg_stat_activity.pid) FROM pg_stat_activity"
+        query_terminate = ("SELECT pg_terminate_backend(pg_stat_activity.pid) FROM pg_stat_activity "
                            "WHERE pg_stat_activity.datname=%(db)s AND pid <> pg_backend_pid()")
     else:
         """ Drop DB connections in Postgres 9.1 and below """
--- a/plugins/modules/postgresql_db.py
+++ b/plugins/modules/postgresql_db.py
@@ -326,11 +326,8 @@ def db_dropconns(cursor, db):
         """ Drop DB connections in Postgres 9.1 and below """
         query_terminate = ("SELECT pg_terminate_backend(pg_stat_activity.procpid) FROM pg_stat_activity"
                            "WHERE pg_stat_activity.datname=%(db)s AND procpid <> pg_backend_pid()")
-    if cursor.connection.server_version < 13000:
-        query_block = ("UPDATE pg_database SET datallowconn = false WHERE datname=%(db)s")
-        query = query_block + ';' + query_terminate
-    else:
-        query = ("DROP DATABASE %(db)s  WITH (FORCE)")
+    query_block = ("UPDATE pg_database SET datallowconn = false WHERE datname=%(db)s")
+    query = query_block + ';' + query_terminate
 
     cursor.execute(query, {'db': db})
 
--- a/plugins/modules/postgresql_db.py
+++ b/plugins/modules/postgresql_db.py
@@ -320,16 +320,18 @@ def db_exists(cursor, db):
 def db_dropconns(cursor, db):
     if cursor.connection.server_version >= 90200:
         """ Drop DB connections in Postgres 9.2 and above """
-        query = """
-        SELECT pg_terminate_backend(pg_stat_activity.pid) FROM pg_stat_activity
-        WHERE pg_stat_activity.datname=%(db)s AND pid <> pg_backend_pid()
-        """
+        query_terminate = ("SELECT pg_terminate_backend(pg_stat_activity.pid) FROM pg_stat_activity"
+                           "WHERE pg_stat_activity.datname=%(db)s AND pid <> pg_backend_pid()")
     else:
         """ Drop DB connections in Postgres 9.1 and below """
-        query = """
-        SELECT pg_terminate_backend(pg_stat_activity.procpid) FROM pg_stat_activity
-        WHERE pg_stat_activity.datname=%(db)s AND procpid <> pg_backend_pid()
-        """
+        query_terminate = ("SELECT pg_terminate_backend(pg_stat_activity.procpid) FROM pg_stat_activity"
+                           "WHERE pg_stat_activity.datname=%(db)s AND procpid <> pg_backend_pid()")
+    if cursor.connection.server_version < 13000:
+        query_block = ("UPDATE pg_database SET datallowconn = false WHERE datname=%(db)s")
+        query = query_block + ';' + query_terminate
+    else:
+        query = ("DROP DATABASE %(db)s  WITH (FORCE)")
+
     cursor.execute(query, {'db': db})
 
 
--- a/plugins/modules/postgresql_db.py
+++ b/plugins/modules/postgresql_db.py
@@ -82,6 +82,11 @@ options:
     type: str
     choices: [ absent, dump, present, rename, restore ]
     default: present
+  force:
+    description:
+    - Used to force drop database
+    type: bool
+    default: False
   target:
     description:
     - File to back up or restore from.
@@ -312,8 +317,26 @@ def db_exists(cursor, db):
     return cursor.rowcount == 1
 
 
-def db_delete(cursor, db):
+def db_dropconns(cursor, db):
+    if cursor.connection.server_version >= 90200:
+        """ Drop DB connections in Postgres 9.2 and above """
+        query = """
+        SELECT pg_terminate_backend(pg_stat_activity.pid) FROM pg_stat_activity
+        WHERE pg_stat_activity.datname=%(db)s AND pid <> pg_backend_pid()
+        """
+    else:
+        """ Drop DB connections in Postgres 9.1 and below """
+        query = """
+        SELECT pg_terminate_backend(pg_stat_activity.procpid) FROM pg_stat_activity
+        WHERE pg_stat_activity.datname=%(db)s AND procpid <> pg_backend_pid()
+        """
+    cursor.execute(query, {'db': db})
+
+
+def db_delete(cursor, db, force=False):
     if db_exists(cursor, db):
+        if force:
+            db_dropconns(cursor, db)
         query = 'DROP DATABASE "%s"' % db
         executed_commands.append(query)
         cursor.execute(query)
@@ -604,6 +627,7 @@ def main():
         tablespace=dict(type='path', default=''),
         dump_extra_args=dict(type='str', default=None),
         trust_input=dict(type='bool', default=True),
+        force=dict(type='bool', default=False)
     )
 
     module = AnsibleModule(
@@ -627,6 +651,7 @@ def main():
     tablespace = module.params['tablespace']
     dump_extra_args = module.params['dump_extra_args']
     trust_input = module.params['trust_input']
+    force = module.params['force']
 
     if state == 'rename':
         if not target:
@@ -713,7 +738,10 @@ def main():
 
         if state == "absent":
             try:
-                changed = db_delete(cursor, db)
+                if force:
+                    changed = db_delete(cursor, db, force=True)
+                else:
+                    changed = db_delete(cursor, db)
             except SQLParseError as e:
                 module.fail_json(msg=to_native(e), exception=traceback.format_exc())
 
--- a/plugins/modules/postgresql_db.py
+++ b/plugins/modules/postgresql_db.py
@@ -629,7 +629,7 @@ def main():
         tablespace=dict(type='path', default=''),
         dump_extra_args=dict(type='str', default=None),
         trust_input=dict(type='bool', default=True),
-        force=dict(type='bool', default=False)
+        force=dict(type='bool', default=False),
     )
 
     module = AnsibleModule(
--- a/plugins/modules/postgresql_db.py
+++ b/plugins/modules/postgresql_db.py
@@ -337,7 +337,7 @@ def db_delete(cursor, db, force=False):
         query = 'DROP DATABASE %(db)s'
         if force:
             if cursor.connection.server_version >= 130000:
-                query = ("DROP DATABASE %(db)s  WITH (FORCE)")
+                query = ('DROP DATABASE "%s" WITH (FORCE)' % db)
             else:
                 db_dropconns(cursor, db)
         executed_commands.append(query)
--- a/plugins/modules/postgresql_db.py
+++ b/plugins/modules/postgresql_db.py
@@ -327,7 +327,7 @@ def db_dropconns(cursor, db):
         query_terminate = ("SELECT pg_terminate_backend(pg_stat_activity.procpid) FROM pg_stat_activity "
                            "WHERE pg_stat_activity.datname=%(db)s AND procpid <> pg_backend_pid()")
     query_block = ("UPDATE pg_database SET datallowconn = false WHERE datname=%(db)s")
-    query = query_block + ';' + query_terminate
+    query = query_block + '; ' + query_terminate
 
     cursor.execute(query, {'db': db})
 
--- a/plugins/modules/postgresql_db.py
+++ b/plugins/modules/postgresql_db.py
@@ -334,7 +334,7 @@ def db_dropconns(cursor, db):
 
 def db_delete(cursor, db, force=False):
     if db_exists(cursor, db):
-        query = 'DROP DATABASE "%(db)s"'
+        query = 'DROP DATABASE %(db)s'
         if force:
             if cursor.connection.server_version >= 130000:
                 query = ("DROP DATABASE %(db)s  WITH (FORCE)")
--- a/plugins/modules/postgresql_db.py
+++ b/plugins/modules/postgresql_db.py
@@ -341,7 +341,7 @@ def db_delete(cursor, db, force=False):
             else:
                 db_dropconns(cursor, db)
         executed_commands.append(query)
-        cursor.execute(query, {'db': db})
+        cursor.execute(query)
         return True
     else:
         return False
--- a/plugins/modules/postgresql_db.py
+++ b/plugins/modules/postgresql_db.py
@@ -324,7 +324,7 @@ def db_dropconns(cursor, db):
                            "WHERE pg_stat_activity.datname=%(db)s AND pid <> pg_backend_pid()")
     else:
         """ Drop DB connections in Postgres 9.1 and below """
-        query_terminate = ("SELECT pg_terminate_backend(pg_stat_activity.procpid) FROM pg_stat_activity"
+        query_terminate = ("SELECT pg_terminate_backend(pg_stat_activity.procpid) FROM pg_stat_activity "
                            "WHERE pg_stat_activity.datname=%(db)s AND procpid <> pg_backend_pid()")
     query_block = ("UPDATE pg_database SET datallowconn = false WHERE datname=%(db)s")
     query = query_block + ';' + query_terminate
--- a/plugins/modules/postgresql_db.py
+++ b/plugins/modules/postgresql_db.py
@@ -334,7 +334,7 @@ def db_dropconns(cursor, db):
 
 def db_delete(cursor, db, force=False):
     if db_exists(cursor, db):
-        query = 'DROP DATABASE %(db)s'
+        query = 'DROP DATABASE "%s"' % db
         if force:
             if cursor.connection.server_version >= 130000:
                 query = ('DROP DATABASE "%s" WITH (FORCE)' % db)
--- a/plugins/modules/postgresql_db.py
+++ b/plugins/modules/postgresql_db.py
@@ -337,9 +337,12 @@ def db_dropconns(cursor, db):
 
 def db_delete(cursor, db, force=False):
     if db_exists(cursor, db):
-        if force:
-            db_dropconns(cursor, db)
         query = 'DROP DATABASE "%s"' % db
+        if force:
+            if cursor.connection.server_version >= 13000:
+                query = ("DROP DATABASE %(db)s  WITH (FORCE)")
+            else:
+                db_dropconns(cursor, db)
         executed_commands.append(query)
         cursor.execute(query)
         return True
--- a/tests/integration/targets/postgresql_db/tasks/main.yml
+++ b/tests/integration/targets/postgresql_db/tasks/main.yml
@@ -42,3 +42,6 @@
   vars:
     file: dbdata.tar
     test_fixture: admin
+
+# Simple test to create and then drop with force
+- import_tasks: manage_database.yml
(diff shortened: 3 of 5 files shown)


changelogs/fragments/drop_db_with_force.yml
```diff
@@ -0,0 +1,3 @@
+postgresql_db name=mydb state=absent fails to remove mydb database if it has active connections.
+
+Added a force option (boolean) to remove active connections first and then remove the database.
```
Andersson007, line 3:
```suggestion
minor_changes:
- postgresql_db - Add the ``force`` boolean option to drop active connections first and then remove the database (https://github.com/ansible-collections/community.postgresql/issues/109).
```

plugins/modules/postgresql_db.py
```diff
@@ -82,6 +82,11 @@
     type: str
     choices: [ absent, dump, present, rename, restore ]
     default: present
+  force:
+    description:
+    - Used to force drop database
```
Andersson007, line 87:
```suggestion
    - Used to forcibly drop a database when the I(state) is C(absent), ignored otherwise.
```

plugins/modules/postgresql_db.py
```diff
@@ -713,7 +738,10 @@ def main():
 
         if state == "absent":
             try:
-                changed = db_delete(cursor, db)
+                if force:
+                    changed = db_delete(cursor, db, force=True)
+                else:
+                    changed = db_delete(cursor, db)
```
Andersson007, line 744:
```suggestion
                changed = db_delete(cursor, db, force)
```

plugins/modules/postgresql_db.py
```diff
@@ -312,8 +317,26 @@ def db_exists(cursor, db):
...
 
 
-def db_delete(cursor, db):
+def db_dropconns(cursor, db):
+    if cursor.connection.server_version >= 90200:
+        """ Drop DB connections in Postgres 9.2 and above """
+        query = """
+        SELECT pg_terminate_backend(pg_stat_activity.pid) FROM pg_stat_activity
+        WHERE pg_stat_activity.datname=%(db)s AND pid <> pg_backend_pid()
+        """
```
Andersson007, line 326:
```suggestion
        query = ("SELECT pg_terminate_backend(pg_stat_activity.pid) FROM pg_stat_activity "
                      "WHERE pg_stat_activity.datname=%(db)s AND pid <> pg_backend_pid()")
```
I'd suggest this format to avoid extra spaces. Same for the query below

Andersson007 (commented):
@asifiqbal hi, thanks for the PR!
Could you also please add an example to the EXAMPLES block.

changelogs/fragments/drop_db_with_force.yml
Andersson007: this should fix the sanity errors

Andersson007:
I wonder if we should the `FORCE` keyword and mention the limitations, e.g. mentioned https://www.enterprisedb.com/postgres-tutorials/postgresql-13-new-feature-drop-database-forcefully

My main concern with the suggested implementation is that on the high load systems there can be many connections from apps established  very fast and it can happen between the `drop_conn` function and the `DROP DATABASE` query  execution and in this case the module will fail.
Should we append the `FORCE` keyword to the `DROP DATABASE` statement when PostgreSQL version is 13 and higher?

Andersson007:
(yeah, it sounds like a corner case and there's no point to drop the database actively used by apps but anyway)

changelogs/fragments/drop_db_with_force.yml
Andersson007: for details about the fragments, refer to https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment

plugins/modules/postgresql_db.py
```diff
@@ -312,8 +317,24 @@ def db_exists(cursor, db):
     return cursor.rowcount == 1
 
 
-def db_delete(cursor, db):
+def db_dropconns(cursor, db):
+    if cursor.connection.server_version >= 90200:
+        """ Drop DB connections in Postgres 9.2 and above """
+        query = ("SELECT pg_terminate_backend(pg_stat_activity.pid) FROM pg_stat_activity "
+                      "WHERE pg_stat_activity.datname=%(db)s AND pid <> pg_backend_pid()")
```
Andersson007, line 324:
```suggestion
        query = ("SELECT pg_terminate_backend(pg_stat_activity.pid) FROM pg_stat_activity "
                      "WHERE pg_stat_activity.datname=%(db)s AND pid <> pg_backend_pid()")
```
the formatting looks a bit strange now:) However if i try to change this, it shows that the first part of the sentence starts and the same position as the second one as exptected.

Ok, we could use the same approach for the query after `else:`

plugins/modules/postgresql_db.py
```diff
@@ -312,8 +317,24 @@ def db_exists(cursor, db):
...
+    else:
+        """ Drop DB connections in Postgres 9.1 and below """
```
Andersson007, line 326:
```suggestion
        """ Drop DB connections in Postgres 9.1 and below """
```
A random thought but should we support very old unsupported versions?

Andersson007: Maybe in this case, i would leave this because it doesn't look complex.

asifiqbal: I agree the code is simple enough to support that old version.

plugins/modules/postgresql_db.py
```diff
@@ -312,8 +317,24 @@ def db_exists(cursor, db):
     return cursor.rowcount == 1
 
 
-def db_delete(cursor, db):
+def db_dropconns(cursor, db):
+    if cursor.connection.server_version >= 90200:
```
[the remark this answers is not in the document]
asifiqbal: I verified. Postgres `9.2.23` shows as 90223 and Postgres `9.1.24` shows as 90124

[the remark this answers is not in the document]
Andersson007: cool, thanks for checking and confirming!
If you'd like to contribute a bugfix, you could create a quick PR to fix `./plugins/modules/postgresql_slot.py`.
There's the check, which is incorrect:
```
if self.cursor.connection.server_version < 96000:
```
What do you think?

plugins/modules/postgresql_db.py
```diff
@@ -312,8 +317,24 @@ def db_exists(cursor, db):
...
+        """ Drop DB connections in Postgres 9.2 and above """
+        query = ("SELECT pg_terminate_backend(pg_stat_activity.pid) FROM pg_stat_activity "
+                 "WHERE pg_stat_activity.datname=%(db)s AND pid <> pg_backend_pid()")
+    else:
+        """ Drop DB connections in Postgres 9.1 and below """
+        query = """
+        SELECT pg_terminate_backend(pg_stat_activity.procpid) FROM pg_stat_activity
+        WHERE pg_stat_activity.datname=%(db)s AND procpid <> pg_backend_pid()
+        """
```
Andersson007, line 330:
```suggestion
        query = ("SELECT pg_terminate_backend(pg_stat_activity.procpid) FROM pg_stat_activity "
                 "WHERE pg_stat_activity.datname=%(db)s AND procpid <> pg_backend_pid()")
```

plugins/modules/postgresql_db.py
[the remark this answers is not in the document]
Andersson007: https://github.com/ansible-collections/community.postgresql/issues/120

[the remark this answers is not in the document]
asifiqbal: > cool, thanks for checking and confirming!
> If you'd like to contribute a bugfix, you could create a quick PR to fix `./plugins/modules/postgresql_slot.py`.
> There's the check, which is incorrect:
> 
> ```
> if self.cursor.connection.server_version < 96000:
> ```
> 
> What do you think?

Right, it should be 90600. I just checked for 9.6.21 and version shows as 90621. I can work the PR. Never done for bug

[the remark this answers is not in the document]
Andersson007: Great thanks! No rush with that as nobody reported the issue yet:)

Andersson007:
hi Asif! I'm going to release the collection within a month. If we have this ready, it would be great. If you need more time, no problem! If you need any help with this or circumstances have changed, please let us know.
Thank you!

mator:
works for me, tested on archlinux current postgresql version:
```
$ pacman -Q postgresql
postgresql 13.3-3
```
git branch of `community.postgresql`:
```
[mator@archlinux community.postgresql.git]$ git branch
* (HEAD detached at asifiqbal/postgres_force_drop)
  main
```
created database:
```
$ sudo createdb mydb
```
put one connection to idle state with `begin` SQL statement:
```
$ echo "SELECT * FROM pg_stat_activity WHERE datname='mydb' " | sudo -u postgres psql
could not change directory to "/home/mator/ansible.dev": Permission denied
 datid | datname |  pid  | leader_pid | usesysid | usename  | application_name | client_addr | client_hostname | client_port |         backend_start         |          xact_start           |          query_start          |         state_change          | wait_event_type | wait_event |        state        | backend_xid | backend_xmin | query  |  backend_type
-------+---------+-------+------------+----------+----------+------------------+-------------+-----------------+-------------+-------------------------------+-------------------------------+-------------------------------+-------------------------------+-----------------+------------+---------------------+-------------+--------------+--------+----------------
 16385 | mydb    | 44407 |            |       10 | postgres | psql             |             |                 |          -1 | 2021-09-10 17:58:19.263529+03 | 2021-09-10 17:58:24.875187+03 | 2021-09-10 17:58:24.875187+03 | 2021-09-10 17:58:24.875352+03 | Client          | ClientRead | idle in transaction |             |              | begin; | client backend
(1 row)
```
used the following playbook:
```
$ cat playbooks/postgresql.yml
---
- hosts: localhost
  become: true
  gather_facts: false
  collections:
  - community.postgresql
  tasks:
  - name: drop database
    postgresql_db:
      name: mydb
      state: absent
      force: true

$ ansible-lint playbooks/postgresql.yml
Failed to guess project directory using git:
INFO     Executing syntax check on playbooks/postgresql.yml (0.84s)
```
playbook execute:
```
[mator@archlinux ansible.dev]$ ansible-playbook playbooks/postgresql.yml

PLAY [localhost] **************************************************************************************************************************************************************

TASK [drop database] **********************************************************************************************************************************************************
changed: [localhost]

PLAY RECAP ********************************************************************************************************************************************************************
localhost                  : ok=1    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```
no `mydb` database afterwards:
```
$ echo '\l mydb' | sudo -u postgres psql
could not change directory to "/home/mator/ansible.dev": Permission denied
                       List of databases
 Name | Owner | Encoding | Collate | Ctype | Access privileges
------+-------+----------+---------+-------+-------------------
(0 rows)
```

Andersson007:
@mator thanks for manual testing!

IIRC Asif and I discussed that would be nice to check the version and if it's 13 or higher, we will use `FORCE` keyword to avoid the situations when there are multiple connections happen quickly that can cause the module's crash (because the new connections can be established between the previous ones were terminated and the `DROP DATABASE` statement execution.
If the version is lower, it will use the mechanism as it uses now.

@asifiqbal @mator @@hunleyd  What do you think? Does it make sense or we can leave things as they are now? As i mentioned, maybe the freaking frequent connections is a corner case and now the implementation is safe enough.

Anyway @asifiqbal could you please apply / criticize my suggestion on formatting and remove `[WIP]` from the issue description?

Thanks!

mator:
does running something like:
```
SQL> update pg_database set datallowconn = false where datname = 'mydb';
```
on a dbs < 13.0 will work? (i.e. restrict new connection to `mydb`, but before dropping all current connections to 'mydb`.

Andersson007:
@mator thanks for the alternatives! all the solutions sound very good. i personally prefer one that works on as many version as possible (I think even 9.4 is still widely used).

cc @asifiqbal @hunleyd your thoughts on @mator 's options ^ (and also please see my previous comment)

plugins/modules/postgresql_db.py
```diff
@@ -312,8 +317,28 @@ def db_exists(cursor, db):
...
+        query = ("DROP DATABASE %(db)s  WITH (FORCE)")
+
+    cursor.execute(query, {'db': db})
+
+
+def db_delete(cursor, db, force=False):
     if db_exists(cursor, db):
+        if force:
+            db_dropconns(cursor, db)
         query = 'DROP DATABASE "%s"' % db
```
Andersson007, line 342:
```suggestion
        query = 'DROP DATABASE "%s"' % db
        if force:
            if cursor.connection.server_version >= 13000:
                query = ("DROP DATABASE %(db)s  WITH (FORCE)")
            else:
                db_dropconns(cursor, db)
```
maybe something like this?

plugins/modules/postgresql_db.py
```diff
@@ -312,8 +317,28 @@ def db_exists(cursor, db):
...
+    else:
+        """ Drop DB connections in Postgres 9.1 and below """
+        query_terminate = ("SELECT pg_terminate_backend(pg_stat_activity.procpid) FROM pg_stat_activity"
+                           "WHERE pg_stat_activity.datname=%(db)s AND procpid <> pg_backend_pid()")
+    if cursor.connection.server_version < 13000:
+        query_block = ("UPDATE pg_database SET datallowconn = false WHERE datname=%(db)s")
+        query = query_block + ';' + query_terminate
+    else:
+        query = ("DROP DATABASE %(db)s  WITH (FORCE)")
+
```
Andersson007, line 334:
```suggestion
    query_block = ("UPDATE pg_database SET datallowconn = false WHERE datname=%(db)s")
    query = query_block + ';' + query_terminate

```
See the proposed change in logic in db_delete.

plugins/modules/postgresql_db.py
```diff
@@ -604,6 +629,7 @@ def main():
         tablespace=dict(type='path', default=''),
         dump_extra_args=dict(type='str', default=None),
         trust_input=dict(type='bool', default=True),
+        force=dict(type='bool', default=False)
```
Andersson007, line 632:
```suggestion
        force=dict(type='bool', default=False),
```

Andersson007 (commented):
@asifiqbal thanks for working on this!

You can try to find `state: absent` in `tests/integration/targets/postgresql_db/tasks/postgresql_db_initial.yml` or in tests/integration/targets/postgresql_db/tasks/postgresql_db_general.yml` to add `force: yes` to one / two tasks

Andersson007:
```
02:10 The full traceback is:
02:10 Traceback (most recent call last):
02:10   File "/tmp/ansible_postgresql_db_payload_mm_bhtwy/ansible_postgresql_db_payload.zip/ansible_collections/community/postgresql/plugins/modules/postgresql_db.py", line 743, in main
02:10   File "/tmp/ansible_postgresql_db_payload_mm_bhtwy/ansible_postgresql_db_payload.zip/ansible_collections/community/postgresql/plugins/modules/postgresql_db.py", line 344, in db_delete
02:10   File "/usr/lib/python3/dist-packages/psycopg2/extras.py", line 142, in execute
02:10     return super(DictCursor, self).execute(query, vars)
02:10 psycopg2.errors.SyntaxError: syntax error at or near "%"
02:10 LINE 1: DROP DATABASE %(db)s  WITH (FORCE)
02:10                       ^
02:10 
02:10 fatal: [testhost]: FAILED! => {
02:10     "changed": false,
02:10     "invocation": {
02:10         "module_args": {
02:10             "ca_cert": null,
02:10             "conn_limit": "",
02:10             "db": "ansible_db",
02:10             "dump_extra_args": null,
02:10             "encoding": "",
02:10             "force": true,
02:10             "lc_collate": "",
02:10             "lc_ctype": "",
02:10             "login_host": "",
02:10             "login_password": "",
02:10             "login_unix_socket": "",
02:10             "login_user": "postgres",
02:10             "maintenance_db": "postgres",
02:10             "name": "ansible_db",
02:10             "owner": "",
02:10             "port": 5432,
02:10             "session_role": null,
02:10             "ssl_mode": "prefer",
02:10             "state": "absent",
```

plugins/modules/postgresql_db.py
```diff
@@ -312,9 +317,29 @@ def db_exists(cursor, db):
...
+def db_delete(cursor, db, force=False):
     if db_exists(cursor, db):
         query = 'DROP DATABASE "%s"' % db
+        if force:
+            if cursor.connection.server_version >= 13000:
+                query = ("DROP DATABASE %(db)s  WITH (FORCE)")
+            else:
+                db_dropconns(cursor, db)
         executed_commands.append(query)
         cursor.execute(query)
```
Andersson007, line 344:
```suggestion
        query = 'DROP DATABASE "%(db)s"'
        if force:
            if cursor.connection.server_version >= 13000:
                query = ("DROP DATABASE %(db)s  WITH (FORCE)")
            else:
                db_dropconns(cursor, db)
        executed_commands.append(query)
        cursor.execute(query, {'db': db})
```

Andersson007:
We can see in CI output that the module is trying to run drop with FORCE when the server version is lower than `13`.
We have to be sure that `if cursor.connection.server_version >= 13000:` is evaluated properly

Andersson007:
Seems that `90504` is greater than `13000` that's why the module runs the block..

Andersson007:
Ah, we should use `130000` https://www.postgresql.org/docs/current/libpq-status.html#LIBPQ-PQSERVERVERSION

Andersson007:
I updated my proposal above to use 130000

Andersson007:
i think the CI failures relate to https://github.com/ansible-collections/overview/issues/45#issuecomment-923507542.
@mator do you have an opportunity to test the PR manually against 9.1, 13, and something between?

Andersson007:
Created an issue https://github.com/ansible-collections/community.postgresql/issues/143 for CI... not tests fail for different reasons in different containers :(
If anyone tests this against 9.1, say, 9.6, and 13 we could merge this with red CI.

Andersson007:
@asifiqbal could you also please
1) remove the `force: yes` from CI where you added it
2) create a simple dedicated task in CI that creates DB
3) drop this DB with `force: yes`

Andersson007:
I found out that we test against 13 with Fedora 34, so no manual tests needed. I'm trying to "fix" them by kicking out all the extra distros:) https://github.com/ansible-collections/community.postgresql/pull/144 After it gets merged, we'll relaunch the tests here.

plugins/modules/postgresql_db.py
```diff
@@ -312,11 +317,31 @@ def db_exists(cursor, db):
...
+    query_block = ("UPDATE pg_database SET datallowconn = false WHERE datname=%(db)s")
+    query = query_block + ';' + query_terminate
+
+    cursor.execute(query, {'db': db})
+
+
+def db_delete(cursor, db, force=False):
     if db_exists(cursor, db):
-        query = 'DROP DATABASE "%s"' % db
+        query = 'DROP DATABASE "%(db)s"'
```
Andersson007, line 337:
```suggestion
        query = 'DROP DATABASE %(db)s'
```

Andersson007: I think this is the reason

plugins/modules/postgresql_db.py
```diff
@@ -312,11 +317,31 @@ def db_exists(cursor, db):
...
+    query_block = ("UPDATE pg_database SET datallowconn = false WHERE datname=%(db)s")
+    query = query_block + ';' + query_terminate
+
+    cursor.execute(query, {'db': db})
+
+
+def db_delete(cursor, db, force=False):
     if db_exists(cursor, db):
-        query = 'DROP DATABASE "%s"' % db
+        query = 'DROP DATABASE %(db)s'
```
Andersson007, line 337:
```suggestion
        query = 'DROP DATABASE "%s"' % db
```

plugins/modules/postgresql_db.py
```diff
@@ -312,11 +317,31 @@ def db_exists(cursor, db):
...
+        if force:
+            if cursor.connection.server_version >= 130000:
+                query = ("DROP DATABASE %(db)s  WITH (FORCE)")
```
Andersson007, line 340:
```suggestion
                query = ('DROP DATABASE "%s" WITH (FORCE)' % db)
```

plugins/modules/postgresql_db.py
```diff
@@ -312,11 +317,31 @@ def db_exists(cursor, db):
...
+            else:
+                db_dropconns(cursor, db)
         executed_commands.append(query)
-        cursor.execute(query)
+        cursor.execute(query, {'db': db})
```
Andersson007, line 344:
```suggestion
        cursor.execute(query)
```

mator:
> @mator do you have an opportunity to test the PR manually against 9.1, 13, and something between?

only after the weekend (read next week)

plugins/modules/postgresql_db.py
```diff
@@ -312,9 +317,29 @@ def db_exists(cursor, db):
     return cursor.rowcount == 1
 
 
-def db_delete(cursor, db):
+def db_dropconns(cursor, db):
+    if cursor.connection.server_version >= 90200:
+        """ Drop DB connections in Postgres 9.2 and above """
+        query_terminate = ("SELECT pg_terminate_backend(pg_stat_activity.pid) FROM pg_stat_activity"
```
Andersson007, line 323:
```suggestion
        query_terminate = ("SELECT pg_terminate_backend(pg_stat_activity.pid) FROM pg_stat_activity "
```

plugins/modules/postgresql_db.py
```diff
@@ -312,9 +317,29 @@ def db_exists(cursor, db):
...
+                           "WHERE pg_stat_activity.datname=%(db)s AND pid <> pg_backend_pid()")
+    else:
+        """ Drop DB connections in Postgres 9.1 and below """
+        query_terminate = ("SELECT pg_terminate_backend(pg_stat_activity.procpid) FROM pg_stat_activity"
```
Andersson007, line 327:
```suggestion
        query_terminate = ("SELECT pg_terminate_backend(pg_stat_activity.procpid) FROM pg_stat_activity "
```

Andersson007:
> > @mator do you have an opportunity to test the PR manually against 9.1, 13, and something between?
> 
> only after the weekend (read next week)

@mator I found out that we test with 13 and 9.1 in CI, so no manual testing is needed, thank you!

plugins/modules/postgresql_db.py
```diff
@@ -312,9 +317,29 @@ def db_exists(cursor, db):
...
+                           "WHERE pg_stat_activity.datname=%(db)s AND procpid <> pg_backend_pid()")
+    query_block = ("UPDATE pg_database SET datallowconn = false WHERE datname=%(db)s")
+    query = query_block + ';' + query_terminate
```
Andersson007, line 330:
```suggestion
    query = query_block + '; ' + query_terminate
```

Andersson007:
@mator but review next week would be much appreciated:)

Andersson007:
:tada:

ready for another review:)

asifiqbal:
> 🎉
> 
> ready for another review:)

Always :-)

Andersson007:
@asifiqbal thanks for the new feature!
@mator thanks for testing!

Andersson007:
`community.postgresql` 1.5.0 has been released and is available to install via the `ansible-galaxy` command.
It also contains these changes!

Issue #133, opened 2021-09-01
postgresql_privs: Add examples of granting privs to functions


Andersson007:
##### SUMMARY
Add examples of granting privs to functions to postgresql_privs module. See https://github.com/ansible-collections/community.postgresql/issues/132  for details.

For newcomers, there's a [quick start guide](https://github.com/ansible/community-docs/blob/main/create_pr_quick_start_guide.rst) how to set up a dev environment and submit a PR . You can also add examples and submit a PR directly via GH GUI.

##### ISSUE TYPE
- Documentation Report

##### COMPONENT NAME
plugin/modules/postgresql_privs.py

Issue #132, opened 2021-08-27
Grant on administrative function pg_start_backup


mnietz:
##### SUMMARY

I'm not able to grant execute rights to pg_start_backup function to any role using postgresql_privs Module. Please find a sample playbook and the exception below.

##### ISSUE TYPE
- Bug Report

##### ANSIBLE VERSION

```
$ ansible --version
ansible 2.9.24
  config file = /home/foo/git/ansible/ansible.cfg
  configured module search path = [u'/home/foo/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.18 (default, Mar  8 2021, 13:02:45) [GCC 9.3.0]
```

##### OS / ENVIRONMENT
Ubuntu 20.04 (Ansible Host). Debian 10 (Postgresq Host) with Postgresql 11

##### STEPS TO REPRODUCE
```
- name: Just a Test
  hosts: pg1
  gather_facts: false

  tasks:

    - name: Grant execute on pg_start_backup to foo
      postgresql_privs:
        port: 5432
        database: postgres
        schema: pg_catalog
        type: function
        objs: "pg_start_backup(text,boolean,boolean)"
        state: present
        privs: EXECUTE
        roles: foo
      become_user: postgres
```

```
postgres=# \df+ pg_catalog.pg_start_backup 
List of functions
-[ RECORD 1 ]-------+-----------------------------------------------------------------------
Schema              | pg_catalog
Name                | pg_start_backup
Result data type    | pg_lsn
Argument data types | label text, fast boolean DEFAULT false, exclusive boolean DEFAULT true
Type                | func
Volatility          | volatile
Parallel            | restricted
Owner               | postgres
Security            | invoker
Access privileges   | postgres=X/postgres
Language            | internal
Source code         | pg_start_backup
Description         | prepare for taking an online backup
```

##### EXPECTED RESULTS

Permission to execute pg_start_backup ist granted to foo.

##### ACTUAL RESULTS

```
he full traceback is:
Traceback (most recent call last):
  File "/tmp/ansible_postgresql_privs_payload_pijzh5e1/ansible_postgresql_privs_payload.zip/ansible/modules/database/postgresql/postgresql_privs.py", line 692, in manipulate_privs
ValueError: not enough values to unpack (expected 2, got 1)

During handling of the above exception, another exception occurred:

[...]

PR #151, opened 2021-10-01
postgresql_privs: Additional information when something goes wrong.

Tas-sos:
##### SUMMARY

In issue #132 ( #133 ) it was reported that when there is an incorrect entry of the parameter `objs` ( or `obj`), it does not arise a more detail message about what is happening.

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME

* plugin/modules/postgresql_privs.py

### DIFF
--- a/plugins/module_utils/database.py
+++ b/plugins/module_utils/database.py
@@ -12,6 +12,7 @@ from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
 import re
+from ansible.module_utils._text import to_native
 
 
 # Input patterns for is_input_dangerous function:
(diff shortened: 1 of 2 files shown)


Andersson007:
@Tas-sos thanks for the PR!
The function is used in many modules, so I suggest moving this `try-except` to the `check_input` function. What do you think?

Tas-sos:
@Andersson007  I thank you for your review :-)
Very good observation! I totally agree with you.

Andersson007 (approved):
LGTM

Andersson007:
@Tas-sos thanks for the contribution!

Issue #161, opened 2021-11-12
Refactor module arguments: Remove `[NO]` prefix


gundalow:
##### SUMMARY
Thanks to @Oloryn and @ssbarnea for reporting this issue and digging into the code.

We need to refactor the module arguments in some of the community.postgresql modules, rather than using the `[NO]SUPERUSER` shorthand as this:
1. isn't how we normally document Ansible module's parameters
2. breaks [vscode-ansible](https://github.com/ansible/vscode-ansible)'s parsing of module arguments.

This was reported in `postgresql_user` parameter `role_attr_flags`, though it likely applies to a few of the postgresql modules

For background please see https://github.com/ansible/vscode-ansible/issues/292

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
postgresql_user

Andersson007:
Looking, thanks for reporting!

PR #162, opened 2021-11-12
postgresql_user: Get rid of wrong documentation

Andersson007:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.postgresql/issues/161

##### ISSUE TYPE
- Docs Pull Request

### DIFF
--- a/plugins/modules/postgresql_user.py
+++ b/plugins/modules/postgresql_user.py
@@ -74,9 +74,8 @@ options:
     - "PostgreSQL user attributes string in the format: CREATEDB,CREATEROLE,SUPERUSER."
     - Note that '[NO]CREATEUSER' is deprecated.
     - To create a simple role for using it like a group, use C(NOLOGIN) flag.
+    - See the full list of supported flags in documentation for your PostgreSQL version.
     type: str
-    choices: [ '[NO]SUPERUSER', '[NO]CREATEROLE', '[NO]CREATEDB',
-               '[NO]INHERIT', '[NO]LOGIN', '[NO]REPLICATION', '[NO]BYPASSRLS' ]
   session_role:
     description:
     - Switch to session role after connecting.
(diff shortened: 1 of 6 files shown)


gundalow:
@Oloryn Here is a proposed fix

Andersson007:
cc @klando could you please take a look?

Andersson007:
cc @tcraxs ^

tcraxs (approved):
LGTM

Oloryn:
So, this essentially just deletes the documentation code that can't currently be handled by vscode-ansible, so that vscode-ansible isn't issuing invalid problems in validation.  Should thought be given to instead replacing it with something that vscode-ansible can validate?  E.g., replace the choices with "[ 'SUPERUSER', 'NOSUPERUSER', 'CREATEROLE', 'NOCREATEROLE'," , etc.?  Plus, there is the issue of role_attr_flags is essentially a list expressed as comma-delimited elements in a string.  Would it make sense to make it actually a list, each of whose elements must match the choices (and deal with backwards compatibility issued that would cause - perhaps by adding a role_attr_flags_list that's actually a list, and down the road deprecate role_attr_flags)?

Andersson007:
@Oloryn thanks for the comment

There are several things worth mentioning:
- it would be a breaking change as you said (as a general rule, it's worth avoiding introducing new options that are not really needed)
- different PG versions can have different set of keywords, so it would add a need to keep tracking keywords (the worst case is getting bug reports from users that new keywords are not supported by the module)
- we had trouble with a similar approach (preset of keywords) in c.mysql and we've recently got read of that (such a relief for maintainers and a good thing for users - they just need to read the doc) 

I personally don't see a good reason to introduce a breaking change and additional maintenance burden  / a new option.
So I would just remove wrong-formatted module's doc and refer users to the official doc for PG versions they use.

klando:
> @Oloryn thanks for the comment
> 
> There are several things worth mentioning:
> 
>     * it would be a breaking change as you said (as a general rule, it's worth avoiding introducing new options that are not really needed)
> 
>     * different PG versions can have different set of keywords, so it would add a need to keep tracking keywords (the worst case is getting bug reports from users that new keywords are not supported by the module)
> 
>     * we had trouble with a similar approach (preset of keywords) in c.mysql and we've recently got read of that (such a relief for maintainers and a good thing for users - they just need to read the doc)
> 
> 
> I personally don't see a good reason to introduce a breaking change and additional maintenance burden / a new option. So I would just remove wrong-formatted module's doc and refer users to the official doc for PG versions they use.

agreed.

Andersson007:
@tcraxs @gundalow @klando @Oloryn thanks for reviewing!

Issue #129, opened 2021-08-20
Better handling of PostgreSQL extension versioning


keithf4:
##### SUMMARY
Better support PostgreSQL extension version handling to support arbitrary naming and downgrades

##### ISSUE TYPE

Feature Request

##### COMPONENT NAME
postgresql_ext module

##### ADDITIONAL INFORMATION
Our organization is still running Ansible 2.9 and we haven't made the full jump to collections and using Galaxy yet. So when testing the installation of PostgreSQL with PostGIS, we ran into this issue with the `unpackaged` extension version causing issues. 

https://github.com/ansible-collections/community.general/pull/1099

We saw that it has been fixed in the community.postgresql collection, but seeing how it was fixed is still a bit concerning for the long term. It is still trying to do a version comparison and, as far as PostgreSQL extensions are concerned, you cannot compare versions like this. This is because the version value and ordering is completely arbitrary, can be any valid alphanumeric value that PG can accept, and there really is no concept of one version being "greater" than another.

How PostgreSQL determines which version is installed and how you can upgrade from one to another is controlled entirely by the update files that an extension author provides. You give one version value followed by another and that provides to PG a valid update path. You can upgrade from one extension version to another, or even downgrade, as long as all the relevant update path files have been provided. For example, for pg_partman, you could literally go from all the way back in 0.1.0 to the current 4.5.1 because there is an unbroken chain (pending any compatiblity issues mentioned in the changelogs of course).

https://github.com/pgpartman/pg_partman/tree/master/updates
```
[...]
-rw-rw-r-- 1 keith keith  94381 Dec 23  2020 pg_partman--4.3.0--4.3.1.sql
-rw-rw-r-- 1 keith keith 131036 May 11  2020 pg_partman--4.3.1--4.4.0.sql
-rw-rw-r-- 1 keith keith 189468 May  3 11:28 pg_partman--4.4.0--4.5.0.sql
-rw-rw-r-- 1 keith keith 188926 May  3 11:28 pg_partman--4.4.1--4.5.0.sql
-rw-rw-r-- 1 keith keith  60613 May 11 14:13 pg_partman--4.5.0--4.5.1.sql
-rw-rw-r-- 1 keith keith  78361 Aug  2 19:10 pg_partman--4.5.1--4.6.0.sql
[...]
```
As far as PG is concerned, it's not technically an upgrade or a downgrade. It's just following the shortest path from the existing version to the target. You can see this from within PostgreSQL as well

```
[...]

PR #163, opened 2021-11-15
Better handling of extension versions

keithf4:
##### SUMMARY
This PR is a bugfix and enhancement for better handling PostgreSQL extensions. The issues surrounding the current handling of extension versions are outlined in the related Issue #129.

Tried to keep it so that there is no change to the current interface or options that the user has to worry about with this update. Only change that the user may see is that a version of "latest" will now always ensure the latest version of an extension is installed whenever the play is run. It did not always do that in all cases before.

The module no longer cares about the actual "version number" of any extension since an extension version is a completely arbitrary string value. What determines which version is "greater" or "less than" another is simply the update paths that the extension author provides, so all version comparison operations have been removed.

For extension updates, this changes the handling of them to simply check if there is a valid update path between the currently installed version and the given target version. This allows supporting of "downgrades" as well if the extension author provides such a method.

Giving a value of "latest" for the extension version works pretty much the same as before, but now it will always attempt to perform an extension update if that value is given. This better supports major version upgrades of PostgreSQL to also include updates to extensions as well if a newer version of the extension is available at that time. For example, when upgrading to PG 13, the version of pg_stat_statements changed to 1.8 to add new columns. If the user had the version of "latest" in their ansible inventory, it would upgrade pg_stat_statements as well as PG if making use of this module.

I'm still not quite familiar with how to get the testing environment for ansible modules set up yet, so I haven't provided updated tests for these changes. I'm attempting to get it figured out, but if someone else could assist with that, that would be great.

I've so far tested these changes with ansible 2.9 and 2.11 myself either by updating the python module manually or by using a custom collections path to the playbook.

Fixes: #129 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
postgresql_ext

### DIFF
--- a/plugins/modules/postgresql_ext.py
+++ b/plugins/modules/postgresql_ext.py
@@ -71,7 +71,10 @@ options:
   version:
     description:
       - Extension version to add or update to. Has effect with I(state=present) only.
-      - If not specified, the latest extension version will be created.
+      - If not specified and extension is not installed in the database,
+        the latest version available will be created.
+      - If extension is already installed, will update to the given version if a valid update
+        path exists.
       - Downgrading is only supported if the extension provides a downgrade path otherwise
         the extension must be removed and a lower version of the extension must be made available.
       - Set I(version=latest) to always update the extension to the latest available version.
@@ -311,7 +314,7 @@ def ext_get_versions(cursor, ext):
 
     cursor.execute(query, params)
 
-    available_versions = [r[0] for r in cursor.fetchall()]
+    available_versions = set(r[0] for r in cursor.fetchall())
 
     if current_version is None:
         current_version = False
@@ -347,7 +350,6 @@ def ext_valid_update_path(cursor, ext, current_version, version):
         params['ver'] = version
 
         cursor.execute(query, params)
-        executed_queries.append(cursor.mogrify(query, params))
         res = cursor.fetchone()
         if res is not None:
             valid_path = True
--- a/plugins/modules/postgresql_ext.py
+++ b/plugins/modules/postgresql_ext.py
@@ -72,7 +72,8 @@ options:
     description:
       - Extension version to add or update to. Has effect with I(state=present) only.
       - If not specified, the latest extension version will be created.
-      - Downgrading is only supported if the extension provides a downgrade path otherwise the extension must be removed and a lower version of the extension must be made available.
+      - Downgrading is only supported if the extension provides a downgrade path otherwise
+        the extension must be removed and a lower version of the extension must be made available.
       - Set I(version=latest) to always update the extension to the latest available version.
     type: str
   trust_input:
--- a/plugins/modules/postgresql_ext.py
+++ b/plugins/modules/postgresql_ext.py
@@ -148,14 +148,13 @@ EXAMPLES = r'''
     cascade: yes
     state: absent
 
-- name: Create extension foo of version 1.2 or update it to that version if it's already
-  created and a valid update path exists community.postgresql.postgresql_ext:
+- name: Create extension foo of version 1.2 or update it to that version if it's already created and a valid update path exists
+  community.postgresql.postgresql_ext:
     db: acme
     name: foo
     version: 1.2
 
-- name: Create the latest available version of extension foo. If already installed,
-  update it to the latest version
+- name: Create the latest available version of extension foo. If already installed, update it to the latest version
   community.postgresql.postgresql_ext:
     db: acme
     name: foo
--- a/plugins/modules/postgresql_ext.py
+++ b/plugins/modules/postgresql_ext.py
@@ -72,9 +72,9 @@ options:
     description:
       - Extension version to add or update to. Has effect with I(state=present) only.
       - If not specified, the latest extension version will be created.
-      - It can't downgrade an extension version.
-        When version downgrade is needed, remove the extension and create new one with appropriate version.
-      - Set I(version=latest) to update the extension to the latest available version.
+      - Version downgrade is only supported if extension provides a downgrade path.
+        Otherwise extension must be removed and lower version of extension must be made available.
+      - Set I(version=latest) to always update the extension to the latest available version.
     type: str
   trust_input:
     description:
@@ -114,6 +114,7 @@ author:
 - Thomas O'Donnell (@andytom)
 - Sandro Santilli (@strk)
 - Andrew Klychkov (@Andersson007)
+- Keith Fiske (@keithf4)
 extends_documentation_fragment:
 - community.postgresql.postgres
 
@@ -147,13 +148,13 @@ EXAMPLES = r'''
     cascade: yes
     state: absent
 
-- name: Create extension foo of version 1.2 or update it if it's already created
+- name: Create extension foo of version 1.2 or update it to that version if it's already created and a valid update path exists
   community.postgresql.postgresql_ext:
     db: acme
     name: foo
     version: 1.2
 
-- name: Assuming extension foo is created, update it to the latest version
+- name: Create latest available version of extension foo. If already installed, update it to the latest version
   community.postgresql.postgresql_ext:
     db: acme
     name: foo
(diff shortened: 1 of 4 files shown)


plugins/modules/postgresql_ext.py
```diff
@@ -72,9 +72,9 @@
     description:
       - Extension version to add or update to. Has effect with I(state=present) only.
       - If not specified, the latest extension version will be created.
-      - It can't downgrade an extension version.
-        When version downgrade is needed, remove the extension and create new one with appropriate version.
-      - Set I(version=latest) to update the extension to the latest available version.
+      - Version downgrade is only supported if extension provides a downgrade path.
+        Otherwise extension must be removed and lower version of extension must be made available.
```
hunleyd, line 76:
```suggestion
      - Downgrading is only supported if the extension provides a downgrade path otherwise the extension must be removed and a lower version of the extension must be made available.
```

plugins/modules/postgresql_ext.py
```diff
@@ -147,13 +148,13 @@
...
 
-- name: Create extension foo of version 1.2 or update it if it's already created
+- name: Create extension foo of version 1.2 or update it to that version if it's already created and a valid update path exists
   community.postgresql.postgresql_ext:
     db: acme
     name: foo
     version: 1.2
 
-- name: Assuming extension foo is created, update it to the latest version
+- name: Create latest available version of extension foo. If already installed, update it to the latest version
```
hunleyd, line 157:
```suggestion
- name: Create the latest available version of extension foo. If already installed, update it to the latest version
```

tcraxs:
@keithf4 One of the extensions checks failed: [see here](https://dev.azure.com/ansible/community.postgresql/_build/results?buildId=29806&view=logs&jobId=073b28e7-9663-5ec9-c2dc-7b282efd4c77&j=073b28e7-9663-5ec9-c2dc-7b282efd4c77&t=b95e671c-cbd3-570c-ec09-f9b58a786974)

It looks like that the assertion is not equal to your new error message ([here](https://dev.azure.com/ansible/community.postgresql/_build/results?buildId=29806&view=logs&j=073b28e7-9663-5ec9-c2dc-7b282efd4c77&t=b95e671c-cbd3-570c-ec09-f9b58a786974&l=7981)) ... and this is your change which triggers that: [line 423](https://github.com/ansible-collections/community.postgresql/pull/163/files#diff-58547ac0bdfdd7d72e8ba12d305173ac989000f4841bbd3c92a9313a88d46ea7R423)

keithf4:
> @keithf4 One of the extensions checks failed: [see here](https://dev.azure.com/ansible/community.postgresql/_build/results?buildId=29806&view=logs&jobId=073b28e7-9663-5ec9-c2dc-7b282efd4c77&j=073b28e7-9663-5ec9-c2dc-7b282efd4c77&t=b95e671c-cbd3-570c-ec09-f9b58a786974)
> 
> It looks like that the assertion is not equal to your new error message ([here](https://dev.azure.com/ansible/community.postgresql/_build/results?buildId=29806&view=logs&j=073b28e7-9663-5ec9-c2dc-7b282efd4c77&t=b95e671c-cbd3-570c-ec09-f9b58a786974&l=7981)) ... and this is your change which triggers that: [line 423](https://github.com/ansible-collections/community.postgresql/pull/163/files#diff-58547ac0bdfdd7d72e8ba12d305173ac989000f4841bbd3c92a9313a88d46ea7R423)
> 
> You have to change [this line](https://github.com/ansible-collections/community.postgresql/blob/81cdbdb90c3fe3526addb5b67ef735be6b5fc961/tests/integration/targets/postgresql_ext/tasks/postgresql_ext_version_opt.yml#L341)

Yes, I did catch that. I have that fixed and am working on some additional python doc changes for the functions and will have that pushed up shortly. Thank you!

keithf4:
I believe I have this PR complete now if anyone else is able to review.

At the time I made it I didn't realize the testing would be done automatically, so took advantage of that to get things fixed and all tests to pass. I believe the existing tests cover all new function usage, so I don't think there's any need for additional tests at this time.

Only lingering concern I had was I had attempted to use the `psycopg2.sql` library to write "safe" sql statements that involve variable identity objects (tables, schemas, etc). However these were failing on CentOS7/U20 since it seems they have psycopg 2.5 by default and that library wasn't added until 2.7. So something to keep in mind for future updates as older OS support is dropped.

https://www.psycopg.org/docs/sql.html

tcraxs (approved):
@keithf4 Thanks for updating this. LGTM

hunleyd (approved):

keithf4:
Reviewing code again, realized I'd created a code path that would never be hit. Fixed!

hunleyd (approved):

hunleyd (approved):

plugins/modules/postgresql_ext.py
```diff
@@ -272,60 +291,71 @@ def ext_get_versions(cursor, ext):
...
-    available_versions = []
+        params['ext'] = ext
+        params['cv'] = current_version
+        params['ver'] = version
 
-    for line in ext_ver_list:
-        if line['version'] == 'unpackaged':
-            continue
+        cursor.execute(query, params)
+        executed_queries.append(cursor.mogrify(query, params))
```
Andersson007, line 350:
```suggestion
        executed_queries.append(cursor.mogrify(query, params))
```
I'm not sure we should log this. This field was introduced to return only queries that modify something, not selects, etc.
I'm not strictly opposed the idea but in this case we, at least, should log all the selects.
I would suggest removing logging of this query because for users, i thinks, it can be important to see what has been changed, not all of the queries that the module executes for its internal purposes.
What do you folks think?

Andersson007 (commented):
@keithf4 thanks for the great improvement! @tcraxs @hunleyd thanks for reviewing!
There are a couple of minor things from me.

plugins/modules/postgresql_ext.py
keithf4: I'd decided to add it, and not other select statements, because I thought it a bit more informative as to why PostgreSQL may be returning that the update path is provided was invalid.
I'm fine with removing it if you'd like though.

plugins/modules/postgresql_ext.py
```diff
@@ -272,60 +291,71 @@ def ext_get_versions(cursor, ext):
...
     # 2. Get available versions:
     query = ("SELECT version FROM pg_available_extension_versions "
              "WHERE name = %(ext)s")
-    cursor.execute(query, {'ext': ext})
-    res = cursor.fetchall()
 
-    available_versions = parse_ext_versions(current_version, res)
+    cursor.execute(query, params)
+
+    available_versions = [r[0] for r in cursor.fetchall()]
```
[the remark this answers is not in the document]
keithf4: I'll readily admit I don't now the intricacies of lists vs sets in python, but I saw you had previously used a list for this variable, so stuck with that to avoid needing possible further code changes elsewhere. Found that quick one-liner for turning a single-value tuple into a list so thought that was a nice simplification. Would you prefer this instead?
```
available_versions = []
for r in cursor.fetchall():
  available_versions.append([r0])
```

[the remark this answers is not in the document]
Andersson007: i wrote this several years ago:)
i meant something like `available_versions = set(r[0] for r in cursor.fetchall())`. Anyway, it doesn't change much, so if you prefer the list, feel free to mark this suggesting as resolved.

plugins/modules/postgresql_ext.py
Andersson007: yeah, i would suggest removing it as we have the path mentioned in the doc

keithf4: Removed. Updated the docs around the version parameter as well.

plugins/modules/postgresql_ext.py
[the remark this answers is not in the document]
keithf4: Changed to the given code example and seems to work without any issues with just that change.

Andersson007 (approved):

Andersson007:
@keithf4 thanks for the great improvement! Would be cool to see more contributions from you!
@tcraxs @hunleyd thanks for reviewing!

I'll try to release the collection today

Andersson007:
Done, 1.6.0. Available via ansible-galaxy now and will be included in the Anbsible package with the next Ansible's patch / minor release.

PR #165, opened 2021-11-22
Add GHA CI for ansible-test sanity

klando:
Edited on https://github.com/ansible-collections/community.postgresql/pull/165#issuecomment-980460927

##### SUMMARY

Setup `ansible-test sanity` with GHA.

Uses upstream projects:

* Action: https://github.com/ansible-community/ansible-test-gh-action
* Action: https://github.com/Data-Bene/ansible-test-versions-gh-action
* Action: https://github.com/dorny/paths-filter
* Worflow: https://github.com/Data-Bene/ansible-collection-gh-workflow

##### COMPONENT NAME

GHA workflow

### DIFF
--- a/.github/workflows/ansible-test-sanity.yml
+++ b/.github/workflows/ansible-test-sanity.yml
@@ -1,0 +1,22 @@
+---
+name: Sanity
+on:
+  pull_request:
+    paths:
+    - 'plugins/**'
+    - '.github/workflows/ansible-test-sanity.yml'
+  push:
+    paths:
+    - 'plugins/**'
+    - '.github/workflows/ansible-test-sanity.yml'
+  schedule:
+  # Every day at 3:14
+  - cron:  '14 3 * * *'
+  workflow_dispatch:
+
+jobs:
+  sanity:
+    uses: data-bene/ansible-collection-gh-workflow/.github/workflows/workflow.yml@release/v1
+    with:
+      fail-fast: false
+      testing-type: sanity
--- a/.github/workflows/ansible-test-sanity.yml
+++ b/.github/workflows/ansible-test-sanity.yml
@@ -16,7 +16,6 @@ on:
 
 jobs:
   sanity:
-    uses: data-bene/ansible-collection-gh-workflow/.github/workflows/workflow.yml@release/v1
+    uses: data-bene/ansible-collection-gh-workflow/.github/workflows/workflow-sanity.yml@release/v1
     with:
       fail-fast: false
-      testing-type: sanity
--- a/.github/workflows/ansible-test-sanity.yml
+++ b/.github/workflows/ansible-test-sanity.yml
@@ -17,5 +17,3 @@ on:
 jobs:
   sanity:
     uses: data-bene/ansible-collection-gh-workflow/.github/workflows/workflow-sanity.yml@release/v1
-    with:
-      fail-fast: false


felixfontein:
The test runs look really strange. It seems that sanity tests for one ansible-core version are run with multiple Python versions? That's something that should not happen. There should be exactly one sanity test run per ansible-core version, which does not specity the `--python` option.

felixfontein:
Also I don't think `--coverage` should be supplied when running sanity tests.

Another comment: it is necessary that the `echo` commands which control grouping also show up in the output? https://github.com/ansible-collections/community.postgresql/runs/4297421739?check_suite_focus=true#step:2:496 That makes the output a lot harder to read then necessary.

klando:
> The test runs look really strange. It seems that sanity tests for one ansible-core version are run with multiple Python versions? That's something that should not happen. There should be exactly one sanity test run per ansible-core version, which does not specity the `--python` option.

Presence of parameters in the command line is managed by the action from https://github.com/ansible-community/ansible-test-gh-action 
The matrix used is: https://github.com/Data-Bene/ansible-test-versions-gh-action/blob/main/versions.yml except that it excludes vars related to docker.

In fact for each ansible version it's tested with 3.8 on controler node. If it should not be specified I can comment in the matrix, assuming the upstream action from ansible-community supports that (but I think by default it will add a `--python 3.9`)

The sanity test are run with full matrix for target-python.

klando:
> Also I don't think `--coverage` should be supplied when running sanity tests.

Managed by upstream action. @webknjaz comments ?

> Another comment: it is necessary that the `echo` commands which control grouping also show up in the output? https://github.com/ansible-collections/community.postgresql/runs/4297421739?check_suite_focus=true#step:2:496 That makes the output a lot harder to read then necessary.

Upstream action is doing that I believe.

webknjaz:
@klando I believe @mattclay said that coverage is necessary in all cases.

felixfontein:
Indeed, I verified that this is useful in some experiments in some collection repos (it adds coverage mostly for external library import guard code). Also it has always been around in the Shippable and AZP configs. I've created https://github.com/ansible-collections/collection_template/pull/34 to enable it in the default workflow for new repos.

felixfontein:
> In fact for each ansible version it's tested with 3.8 on controler node. If it should not be specified I can comment in the matrix, assuming the upstream action from ansible-community supports that (but I think by default it will add a `--python 3.9`)

It provides an explicit example: https://github.com/ansible-community/ansible-test-gh-action#usage
```.yaml
- name: Perform sanity testing with ansible-test
  uses: ansible-community/ansible-test-gh-action@release/v1
  with:
    ansible-core-version: stable-2.11
    testing-type: sanity
```

> The sanity test are run with full matrix for target-python.

For smaller repos, there's no need to do that either. Simply don't give ansible-test a (target) Python version, and it will automatically run unit tests for all Python versions it supports (when using `--docker default`).

(Also the non-target Python version shouldn't be specified IMO, so that ansible-community/ansible-test-gh-action can pick a version that works for the current ansible-core branch. After all that Python version is completely irrelevant for the tests, since the tests are run inside a Docker container.)

klando:
> For smaller repos, there's no need to do that either. Simply don't give ansible-test a (target) Python version, and it will automatically run unit tests for all Python versions it supports (when using `--docker default`).

then it's a longer job, with larger output.

I noticed that I need to exclude:

```
target-python-exclude:
- ansible-core-version: stable-2.9
  target-python-version: 3.9
- ansible-core-version: stable-2.9
  target-python-version: 3.10
- ansible-core-version: stable-2.10
  target-python-version: 3.10
- ansible-core-version: stable-2.11
  target-python-version: 3.10
- ansible-core-version: devel
  target-python-version: 2.6
```

but ansible-test knows that already... is it possible to get this from ansible-test ? (`ansible-test --list-target-python-versions` or the like) ?


> (Also the non-target Python version shouldn't be specified IMO, so that ansible-community/ansible-test-gh-action can pick a version that works for the current ansible-core branch. After all that Python version is completely irrelevant for the tests, since the tests are run inside a Docker container.)

Good for me, will update.

felixfontein:
> > For smaller repos, there's no need to do that either. Simply don't give ansible-test a (target) Python version, and it will automatically run unit tests for all Python versions it supports (when using `--docker default`).
> 
> then it's a longer job, with larger output.

I don't think that's a problem. Also it only goes through the Python until the first one fails, then it stops. So if you have output you have to look at, it's usually not that long, and the interesting parts are always at the end.

> I noticed that I need to exclude:
> 
> ```
> target-python-exclude:
> - ansible-core-version: stable-2.9
>   target-python-version: 3.9
> - ansible-core-version: stable-2.9
>   target-python-version: 3.10
> - ansible-core-version: stable-2.10
>   target-python-version: 3.10
> - ansible-core-version: stable-2.11
>   target-python-version: 3.10
> - ansible-core-version: devel
>   target-python-version: 2.6
> ```
> 
> but ansible-test knows that already... is it possible to get this from ansible-test ? (`ansible-test --list-target-python-versions` or the like) ?

No idea how to obtain it automatically, but that's another reason why I would simply let `ansible-test unit` handle this.

klando:
I've updated reusable workflow to not create a job matrix with python targets as it is handled by ansible-test already.
I've limited the reusable workflow to sanity only, thus the name changed from `workflow.yml` to  `workflow-sanity.yml`

ansible-test change detection is not implemented in uptsream ansible-test-gh-action so I keep doing it in the reusabel workflow, can be changed if/when the upstream action is updated.

webknjaz:
I don't like that it's pointing at a moving target rather than a tag or a sha hash.

klando:
> I don't like that it's pointing at a moving target rather than a tag or a sha hash.

it was just for testing purpose and make sure everybody is happy with that.

.github/workflows/ansible-test-sanity.yml
```diff
@@ -0,0 +1,21 @@
...
+  schedule:
+  # Every day at 3:14
+  - cron:  '14 3 * * *'
+  workflow_dispatch:
+
+jobs:
+  sanity:
+    uses: data-bene/ansible-collection-gh-workflow/.github/workflows/workflow-sanity.yml@devel
+    with:
+      fail-fast: false
```
felixfontein, line 21:
```suggestion
      fail-fast: true
```
There's usually no reason not to do fail fast.

.github/workflows/ansible-test-sanity.yml
```diff
@@ -0,0 +1,21 @@
...
+    - 'plugins/**'
+    - '.github/workflows/ansible-test-sanity.yml'
+  schedule:
+  # Every day at 3:14
+  - cron:  '14 3 * * *'
+  workflow_dispatch:
+
+jobs:
+  sanity:
+    uses: data-bene/ansible-collection-gh-workflow/.github/workflows/workflow-sanity.yml@devel
```
felixfontein, line 19: Since there is no matrix: how does it determine the minimal and maximal supported stable branches?

The minimum can probably be extracted from `meta/runtime.yml`'s `requires_ansible`.

The maximum should be a parameter here, and there should be a boolean flag for whether `devel` is included or not.

klando:
##### SUMMARY

Setup `ansible-test sanity` with GHA.

This version depends only on actions and workflow which can be committed to the ansible community. It runs less jobs, and let ansible-test manage the python version and target python version to use.

It runs on all ansible stable version and devel, for all targets by default (the reusable workflow supports targets parameter but nothing automatic is done in this area for keep it simple).

Uses upstream projects:

* Action: https://github.com/ansible-community/ansible-test-gh-action
* Action: https://github.com/Data-Bene/ansible-test-versions-gh-action
* Worflow: https://github.com/Data-Bene/ansible-collection-gh-workflow





##### COMPONENT NAME

GHA workflow

klando:
I left a schedule for the workflow, it can be removed. Initially I added it for external updates (new ansible minor version, etc.)

.github/workflows/ansible-test-sanity.yml
klando: I've reduced complexity but it's still very easy to add default by providing inputs parameters to the workflow.
The action currently works with a static yml file to keep versions, but it can be improved, and in fact this `meta/runtime.yml` is of interest for more automatic setup, tks for pointing to it.

Andersson007:
Can we simply have something like the following instead of having all that projects involved? (My view is described in https://github.com/ansible-collections/community.postgresql/pull/166#issuecomment-982362827)
```
  sanity:
    name: Sanity (Ⓐ${{ matrix.ansible }})
    strategy:
      matrix:
        ansible:
          # It's important that Sanity is tested against all stable-X.Y branches
          # Testing against `devel` may fail as new tests are added.
          - stable-2.9
          - stable-2.10
          - stable-2.11
          - stable-2.12
          - devel
    runs-on: ubuntu-latest
    steps:

      # ansible-test requires the collection to be in a directory in the form
      # .../ansible_collections/${{env.NAMESPACE}}/${{env.COLLECTION_NAME}}/

      - name: Check out code
        uses: actions/checkout@v2
        with:
          path: ansible_collections/${{env.NAMESPACE}}/${{env.COLLECTION_NAME}}

      - name: Set up Python
        uses: actions/setup-python@v2
        with:
          # it is just required to run that once as "ansible-test sanity" in the docker image
          # will run on all python versions it supports.
          python-version: 3.8

      # Install the head of the given branch (devel, stable-2.10)
      - name: Install ansible-core (${{ matrix.ansible }})
        run: pip install https://github.com/ansible/ansible/archive/${{ matrix.ansible }}.tar.gz --disable-pip-version-check

      # run ansible-test sanity inside of Docker.
      # The docker container has all the pinned dependencies that are required
      # and all python versions ansible supports.
      - name: Run sanity tests
        run: ansible-test sanity --docker -v --color --coverage
        working-directory: ./ansible_collections/${{env.NAMESPACE}}/${{env.COLLECTION_NAME}}

        # ansible-test support producing code coverage date
      - name: Generate coverage report
        run: ansible-test coverage xml -v --requirements --group-by command --group-by version
        working-directory: ./ansible_collections/${{env.NAMESPACE}}/${{env.COLLECTION_NAME}}

      # See the reports at https://codecov.io/gh/ansible-collections/community.dns
      - uses: codecov/codecov-action@v2
        with:
          fail_ci_if_error: false
```

mattclay:
> @klando I believe @mattclay said that coverage is necessary in all cases.

Yes, if code coverage is desired, then the `--coverage` option should be used for all test types, including sanity tests.

Andersson007:
closing to keep the tracker clean, thanks everyone!

PR #179, opened 2022-01-06
Use vendored version of distutils.version instead of the deprecated Python standard library distutils

Andersson007:
##### SUMMARY
https://github.com/ansible-collections/community.postgresql/pull/178

### DIFF
--- a/plugins/module_utils/version.py
+++ b/plugins/module_utils/version.py
@@ -8,14 +8,9 @@
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
-from ansible.module_utils.six import raise_from
+# Once we drop support for Ansible 2.9 and ansible-base 2.10, we can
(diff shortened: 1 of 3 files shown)


plugins/module_utils/version.py
```diff
@@ -8,14 +8,9 @@
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
-from ansible.module_utils.six import raise_from
+# Once we drop support for Ansible 2.9 and ansible-base 2.10, we can
```
felixfontein, line 11:
```suggestion
# Once we drop support for Ansible 2.9, ansible-base 2.10, and ansible-core 2.11, we can
```

felixfontein (approved):

Andersson007:
Thanks @felixfontein !

PR #220, opened 2022-03-29
[WIP] Add psycopg3 support. Please do NOT review it now.

Andersson007:
##### SUMMARY
This is a very rough draft. Please DO NOT review it now

**UPDATE 2022-03-31**: I stumbled upon a really big issue https://www.psycopg.org/psycopg3/docs/basic/from_pg2.html#server-side-binding - it means that parametrized **DDL** queries (like ALTER USER, CREATE DATABASE, etc.) are NOT supported any more...:( I'm afraid psycopg3 support will require too heavy refactoring comparable to writing alternative implementations for all of the modules... The suggested in the doc workaround with using psycopg.sql.SQL() and format() will not work easily because `format()` does not understand named arguments like in "VALID UNTIL %(expires)s"... Huge issue.. The safest way would be the alternative implementations (at least of many functions) mentioned but who wants to invest such a lot of time (my estimate is, at least, weeks of full time work)...

Done with:
- postgresql_query (there are several implementation questions that need to be discussed later)
- postgresql_set

Roadmap:
- [ ] Make all of the modules working with psycopg3 (use integration tests adjusted to use psycopg3 locally)
- [ ] Try to refactor modules accordingly now in a part that relates to psycopg2, submit separate PRs
- [ ] Check that all the modules also work with psycopg2 (revert related changes in CI, run them locally)
- [ ] Discuss things that need to be discussed (use comments to mark such places)
- [ ] Complete / change implementation based on discussion results
- [ ] Find a way to add testing against both psycopg2 and psycopg3
- [ ] Think of a corresponding major release

Alternative approach:
- [ ] Introduction of non-breaking psycopg3 support to modules one by one without major releases. Question: what to do with CI?

### DIFF
--- a/plugins/module_utils/postgres.py
+++ b/plugins/module_utils/postgres.py
@@ -17,10 +17,14 @@ from decimal import Decimal
 
 psycopg2 = None  # This line needs for unit tests
 try:
-    import psycopg2
+    import psycopg as psycopg2
     HAS_PSYCOPG2 = True
 except ImportError:
-    HAS_PSYCOPG2 = False
+    try:
+        import psycopg2
+        HAS_PSYCOPG2 = True
+    except ImportError:
+        HAS_PSYCOPG2 = False
 
 from ansible.module_utils.basic import missing_required_lib
 from ansible.module_utils._text import to_native
@@ -56,7 +60,7 @@ def ensure_required_libs(module):
         module.fail_json(msg='psycopg2 must be at least 2.4.3 in order to use the ca_cert parameter')
 
 
-def connect_to_db(module, conn_params, autocommit=False, fail_on_conn=True):
+def connect_to_db(module, conn_params, autocommit=False, fail_on_conn=True, row_factory=None):
     """Connect to a PostgreSQL database.
 
     Return a tuple containing a psycopg2 connection object and error message / None.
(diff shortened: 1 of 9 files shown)


Andersson007:
@hunleyd any thoughts on the UPDATE 2022-03-31 in the description?

Andersson007:
@hunleyd thanks for the feedback!
> is there an immediate need to switch to v3? or do we have some time?
- I don't think so. I also haven't found any info about `psycopg2`'s EOL date. I think if psycopg3 is 3.0.* now, we will have a couple of years (i hope).

> i wonder if we should open some issues w/ the psycopg3 maintainer to discuss your findings and see what they suggest and maybe future releases can be tweaked to lower the burden of conversion for us.
- Great idea, we could start with opening an issue about parameterized DDL statements. I'll investigate a bit more and will create the issue. Thanks!

Andersson007:
@hunleyd how about the following:
===========================

Hello and thank you for working on the adapter of version 3.

@hunleyd and I are maintainers of the [community.postgresql](https://github.com/ansible-collections/community.postgresql) Ansible collection which has been a part of Ansible package for many years and currently uses `psycopg2`.

We are trying to add `psycopg3` support to the collection.

**Question 1**: We haven't found EOL date of `psycopg2`, does it exists? If not yet, are there any related plans?

Most of the statements are DDL in our modules.

**Question 2**: Are there any plans to introduce the parameterized DDL statement execution in the future or there are technical restrictions for that?

Context: In most cases DDL statements in the collection are combined dinamically,
so the result passed to execution can be different in the same function, for example
(I'll replace the statement variable and parameters with their possible values):
- cursor.execute("ALTER USER ... PASSWORD %(password)s", {"password": "lsdjfs", "expires": "2022-02-01"})
- cursor.execute("ALTER USER ... PASSWORD %(password)s VALID UNTIL %(expires)s", {"password": "lsdjfs", "expires": "2022-02-01"})
- cursor.execute("ALTER USER ... VALID UNTIL %(expires)s", {"password": "lsdjfs", "expires": "2022-02-01"})

As we can see it uses, as i call it, the named args format, for example, `%(password)s` and the dictionary containing parameters, and in some runs all of the dict elements can be used and in some runs only a part.

As said in https://www.psycopg.org/psycopg3/docs/basic/from_pg2.html#server-side-binding , parameterized DDL statements
are not supported any more.

The doc suggests using `psycopg.sql.SQL()` and `format()` BUT it doesn't seem to support the `%(key)s` format, but only `{key}` and others similar to string.format().

**Question 3**: Is that correct that `format()` does not understand things like `%(key)s`, only `{key}`? If yes, this is a huge issue for us (see the above examples) as it will require to write an alternative implementation for almost every function to support `psycopg3` (tens of thousands lines of code.. + same for unit tests, etc., which is tough for 2 folks)

**Question 4**: If the things mentioned in questions 2 and 3 are impossible to change because of `Server-side binding` or whatever, is there a way (or plans to introduce it) for users to specify which binding (server- or client-side) the driver will use?

Looking forward to any feedback, thank you!

Andersson007:
We should wait for https://github.com/psycopg/psycopg/issues/101 gets implemented.

PR #285, opened 2022-06-03
fix: Cast connparams to allow idempotence (#280)

### LINKED ISSUE #280: postgresql_subscription module is not idempotent
##### ISSUE TYPE
- Bug Report

##### ANSIBLE VERSION

```paste below
ansible [core 2.11.6] 
  config file = /home/jgalarza/playbook/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.8/dist-packages/ansible
  ansible collection location = /home/jgalarza/playbook/collections
  executable location = /usr/local/bin/ansible
  python version = 3.8.10 (default, Mar 15 2022, 12:22:08) [GCC 9.4.0]
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```paste below
community.postgresql 2.1.4
```

##### CONFIGURATION

```paste below
COLLECTIONS_PATHS(/home/jgalarza/playbook/ansible.cfg) = ['/home/jgalarza/bidb-rds->
DEFAULT_FORKS(/home/jgalarza/playbook/ansible.cfg) = 20
DEFAULT_REMOTE_USER(/home/jgalarza/playbook/ansible.cfg) = ansible
DEFAULT_ROLES_PATH(/home/jgalarza/playbook/ansible.cfg) = ['/home/jgalarza/bidb-rds>
DEPRECATION_WARNINGS(/home/jgalarza/playbook/ansible.cfg) = False
HOST_KEY_CHECKING(/home/jgalarza/playbook/ansible.cfg) = False
```

##### OS / ENVIRONMENT

Ubuntu 20.04.4 LTS

##### STEPS TO REPRODUCE

```yaml
- name: Start subscription
  community.postgresql.postgresql_subscription:
    port: "{{ db_port }}"
    login_host: "{{ db_host }}"
    login_user: "{{ db_u }}"
    login_password: "{{ db_p }}"
    db: "{{ item.1.PGDATABASE }}"
    name: "{{ item.1.PGDATABASE }}__mypub"
    state: present
    publications: mypub
    owner: postgres
    connparams:
      host: "{{ hostvars[item.1.source.host].ansible_ssh_host }}"
      port: "{{ item.1.source.PGPORT }}"
      user: repuser
      dbname: "{{ item.1.source.PGDATABASE }}"
  register: sub_creation
  with_subelements:
    - "{{ pgcluster }}"
    - databases
  tags: [debug, database]
```

##### EXPECTED RESULTS

The task should be idempotent

##### ACTUAL RESULTS

The task always reports a change and run the following query:
```SQL
ALTER SUBSCRIPTION mydb__mypub CONNECTION 'host=x.x.x.x po
rt=5432 user=repuser dbname=mydb'
```
Looks like the module always reports a change in the connection details. 

The issue may be related to this [line](https://github.com/ansible-collections/community.postgresql/blob/main/plugins/modules/postgresql_subscription.py#L385)

```paste below

```

jegj:
##### SUMMARY

Fixes #280. `postgresql_subscription` module is not idempotent because the port is cast to `int`  in the following chunk of code( [335](https://github.com/ansible-collections/community.postgresql/blob/639197ead257c1b4c63627d035996790e904cf6e/plugins/modules/postgresql_subscription.py#L335)):

```py
                try:
                    self.attrs['conninfo'][tmp[0]] = int(tmp[1])
                except ValueError:
                    self.attrs['conninfo'][tmp[0]] = tmp[1]
```
On line [385](https://github.com/ansible-collections/community.postgresql/blob/639197ead257c1b4c63627d035996790e904cf6e/plugins/modules/postgresql_subscription.py#L385) the if condition always return false  because variable `connparams` is a dict of string
```py
            if connparams != self.attrs['conninfo']:
                changed = self.__set_conn_params(convert_conn_params(connparams),
                                                 check_mode=check_mode)
```

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
postgresql_subscription

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/285-postgresql_subscription_fix_idempontece.yml
+++ b/changelogs/fragments/285-postgresql_subscription_fix_idempontece.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - postgresql_subscription - Fix idempotence by casting the connparams dict variable
+  - postgresql_subscription - fix idempotence by casting the ``connparams`` dict variable (https://github.com/ansible-collections/community.postgresql/issues/280).
--- a/changelogs/fragments/285-postgresql_subscription_fix_idempontece.yml
+++ b/changelogs/fragments/285-postgresql_subscription_fix_idempontece.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - postgresql_subscription - Fix idempotence by casting the connparams dict variable
--- a/plugins/modules/postgresql_subscription.py
+++ b/plugins/modules/postgresql_subscription.py
@@ -331,7 +331,10 @@ class PgSubscription():
         if subscr_info.get('subconninfo'):
             for param in subscr_info['subconninfo'].split(' '):
                 tmp = param.split('=')
-                self.attrs['conninfo'][tmp[0]] = tmp[1]
+                try:
+                    self.attrs['conninfo'][tmp[0]] = int(tmp[1])
+                except ValueError:
+                    self.attrs['conninfo'][tmp[0]] = tmp[1]
 
         return True
 
--- a/plugins/modules/postgresql_subscription.py
+++ b/plugins/modules/postgresql_subscription.py
@@ -264,7 +264,7 @@ def convert_subscr_params(params_dict):
 
     return ', '.join(params_list)
 
- def cast_connparams(connparams):
+def cast_connparams(connparams):
     """Cast the passed connparams dictionary
 
     Returns:
--- a/plugins/modules/postgresql_subscription.py
+++ b/plugins/modules/postgresql_subscription.py
@@ -264,6 +264,7 @@ def convert_subscr_params(params_dict):
 
     return ', '.join(params_list)
 
+
 def cast_connparams(connparams):
     """Cast the passed connparams dictionary
 
--- a/plugins/modules/postgresql_subscription.py
+++ b/plugins/modules/postgresql_subscription.py
@@ -331,10 +331,7 @@ class PgSubscription():
         if subscr_info.get('subconninfo'):
             for param in subscr_info['subconninfo'].split(' '):
                 tmp = param.split('=')
-                try:
-                    self.attrs['conninfo'][tmp[0]] = int(tmp[1])
-                except ValueError:
-                    self.attrs['conninfo'][tmp[0]] = tmp[1]
+                self.attrs['conninfo'][tmp[0]] = tmp[1]
 
         return True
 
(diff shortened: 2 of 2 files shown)


hunleyd:
thank you very much for the PR @jegj . Can you fix the failing tests and include a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#collection-changelog-fragments)?

Andersson007:
@jegj thanks for the fix!
i'm not sure if the patch is safe
Look (this is from our CI):
```
01:31    "invocation": {
01:31             "connparams": {
01:31                 "dbname": "acme_db",
01:31                 "host": "127.0.0.1",
01:31                 "password": "alsdjfKJKDf1#",
01:31                 "port": 5433,
01:31                 "user": "logical_replication"
01:31             },
```
The port can be passed as `int`. How does you playbook look, do you pass it as `"5433"` probably?
Then yeah it's a bug (which appear when we pass it as `string`) but if we merge the PR, it'll break things for users who pass the port as `int`.
In this case we should fix it by doing the same for `connparams` as we do for `self.attrs['conninfo']` (with that `try-except`).
I suggest adding a function which will do it for `connparams`.
What do you think?

jegj:
@Andersson007 I pass the port as `int`. I tried with `string` same problem. But you are right,  that is a better solution. I will add that block code

Andersson007:
> @Andersson007 I pass the port as `int`. I tried with `string`, same problem. But you are right, that is a better solution. I will add that block code

@jegj strange, i should check it, maybe it's converted to string internally somehow but ansible output shows as it's int..

jegj:
@Andersson007 I have added the function but still getting errors on CI. I will take a deep look and see what's the problem

Andersson007:
@jegj just a fyi, you can run the integration tests locally before pushing, see https://docs.ansible.com/ansible/devel/community/create_pr_quick_start.html. It'll save a bit of time as it's faster than waiting for results here

changelogs/fragments/285-postgresql_subscription_fix_idempontece.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - postgresql_subscription - Fix idempotence by casting the connparams dict variable
```
Andersson007, line 2:
```suggestion
  - postgresql_subscription - fix idempotence by casting the ``connparams`` dict variable (https://github.com/ansible-collections/community.postgresql/issues/280).
```
To make it looking consistent with other entries

Andersson007 (approved):

Andersson007:
@jegj thanks very much for the contribution! We'll be happy to see more by you and help whenever needed!

jegj:
Happy to help!

PR #282, opened 2022-06-02
Update postgresql_privs.py to include `ALL x IN SCHEMA` functionality for PostgreSQL versions 9.x +

jchancojr:
Alter postgresql_privs.py to include `ALL x IN SCHEMA` functionality for PostgreSQL versions 9.0.0 or greater. We still allow versions 8.4.22 to `GRANT` to all SCHEMA, but this is done by building a list and granting to the objects on that list individually. 

Fixes #269 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
postgresql_privs.py

PR #287, opened 2022-06-05
backport 282

hunleyd:
backport #282


hunleyd:
does that look correct @jchancojr and @Andersson007 ?

Andersson007 (approved):
@hunleyd looks same to me

Issue #276, opened 2022-05-16
formalized support matrix for PG versions


hunleyd:
##### SUMMARY

The collection should adopt a matrix of supported PG versions. Right now, there's no formal declaration of what PG versions are supported and the pipeline tests _a few_ different versions but the answer to 'what versions of PG does this collection support?' is essentially 'all of them? ¯\\\_(ツ)\_/¯'

As noted in [this comment](https://github.com/ansible-collections/community.postgresql/issues/269#issuecomment-1126783632), we have places in the code that are marked as 'we do it this way because versions older than X don't support FOO' and yet, we don't actually test those older versions to ensure it works and we can't make optimizations in the code for fear of breaking old PG releases.

##### PROPOSAL

The PostgreSQL Global Development Group (PGDG) officially support every major version of PG for 5 years according to their [version policy](https://www.postgresql.org/support/versioning/). I propose that this collection follow the currently supported PGDG releases.

An illustrative example based on this year's upcoming PG 15 release:

| Year | Matrix |
| -- | -- |
| 2022 | 15,14,13,12,11 |
| 2023 | 16,15,14,13,12 |
| 2024 | 17,16,15,14,13 |
| 2025 | 18,17,16,15,14 |
| 2026 | 19,18,17,16,15 |
| 2027 | 20,19,18,17,16 |

The table above shows the PG versions that would be tested via the pipeline to ensure compatibility. We would release a major version of the collection each year when the pipeline matrix is updated (maybe we bump the collection version number to match the PG release being supported? So adding support for PG 15 would give us `community.postgresql-15.0.0`. Food for thought.)

##### REASONING

This would allow us to drop/refactor old code on a regular basis and be free to take advantage of new PG features more easily. End users would gain a clearer understanding of what PG versions are tested in a given release, and if they need to continue supporting an older release, they could simply use a prior release that is known to support that version (e.g. in 2027 if they still needed PG 15 support they could simply use the collection release from 2026).

keithf4:
This sounds good. I've followed a similar support pattern with pg_partman. I didn't go out of my way to make the extension not work with older versions; it still technically works back to PG 9.4 but I officially only support 10+.  But if a feature or bugfix came out in PG that was not supported by a version older than that window, then I wouldn't hesitate to implement that feature and therefore break that backward compatibility at that time.

pgguru:
[...]

PR #290, opened 2022-06-12
tests: setup five pg versions for testing

hunleyd:
##### SUMMARY

Alter the text matrix to specify the PG versions as discussed in #276

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/tests/integration/targets/setup_postgresql_db/tasks/dummy_ext.yml
+++ b/tests/integration/targets/setup_postgresql_db/tasks/dummy_ext.yml
@@ -1,0 +1,27 @@
+---
+- name: copy control file for dummy ext
+  copy:
+    src: dummy.control
+    dest: /usr/share/postgresql/{{ pg_ver }}/extension/dummy.control
+    mode: '0444'
+
+- name: copy version files for dummy ext
+  copy:
+    src: '{{ item }}'
+    dest: /usr/share/postgresql/{{ pg_ver }}/extension/{{ item }}
+    mode: '0444'
+  with_items:
+  - dummy--0.sql
+  - dummy--1.0.sql
+  - dummy--2.0.sql
+  - dummy--3.0.sql
+
+- name: add update paths
+  file:
+    path: /usr/share/postgresql/{{ pg_ver }}/extension/{{ item }}
+    mode: '0444'
+    state: touch
+  with_items:
+  - dummy--0--1.0.sql
+  - dummy--1.0--2.0.sql
+  - dummy--2.0--3.0.sql
--- a/tests/integration/targets/setup_postgresql_db/tasks/install_deps.yml
+++ b/tests/integration/targets/setup_postgresql_db/tasks/install_deps.yml
@@ -1,0 +1,14 @@
+---
+- name: disable PG module (RH8+)
+  command: dnf -y module disable postgresql
+  when:
+    - ansible_facts.os_family == 'RedHat'
+    - ansible_facts.distribution_major_version is version('8', '>=')
+
+- name: install dependencies for postgresql test
+  package:
+    name: '{{ postgresql_package_item }}'
+    state: present
+  with_nested: '{{ postgresql_packages }}'
+  loop_control:
+    loop_var: postgresql_package_item
--- a/tests/integration/targets/setup_postgresql_db/tasks/main.yml
+++ b/tests/integration/targets/setup_postgresql_db/tasks/main.yml
@@ -25,15 +25,14 @@
       paths:
       - '{{ role_path }}/vars'
 
-- name: Make sure the dbus service is enabled under systemd
-  shell: systemctl enable dbus || systemctl enable dbus-broker
-  ignore_errors: yes
-  when: ansible_service_mgr == 'systemd' and ansible_distribution == 'Fedora'
-
-- name: Make sure the dbus service is started under systemd
-  systemd:
-    name: dbus
-    state: started
+- block:
+    - name: Make sure the dbus service is enabled under systemd
+      shell: systemctl enable dbus || systemctl enable dbus-broker
+      ignore_errors: yes
+    - name: Make sure the dbus service is started under systemd
+      systemd:
+        name: dbus
+        state: started
   when: ansible_service_mgr == 'systemd' and ansible_distribution == 'Fedora'
 
 - name: Kill all postgres processes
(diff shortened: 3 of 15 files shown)


hunleyd:
hey @Andersson007 .. this is nowhere near complete, but i have questions for you:
1. does this approach make sense to you (installing on 5 versions in the test instance at once)?
2. can i remove `tests/integration/targets/*/vars/*` for OSes that are not listed in the Azure pipeline?

Those will get me started. I'm sure I'll be back with more :)

Andersson007:
> hey @Andersson007 .. this is nowhere near complete, but i have questions for you:
> 
> 1. does this approach make sense to you (installing all 5 versions in the test instance at once)?
> 2. can i remove `tests/integration/targets/*/vars/*` for OSes that are not listed in the Azure pipeline?
> 
> Those will get me started. I'm sure I'll be back with more :)

@hunleyd , first of all, thank you for working on this!

> 1. does this approach make sense to you (installing all 5 versions in the test instance at once)?

I'm not sure i fully understand what the implementation plan is.
Could you please explain it?
IMO, ideally it should look like in https://github.com/ansible-collections/community.mysql/pull/370.
If you open the test section, you'll see entries like `Integration (Python: 3.8, Ansible: stable-2.12, MySQL: mariadb_10.8.3, Connector: mysqlclient==2.0.1`.
- So there we see separate tasks for each version of MySQL and the connector. I think it's very convenient to have them separate. 
- I'm not sure if this is achievable in AZP.
- GHA is not an option if we don't find a way to run the tests using diff, i.e. only tests for changed modules must run, not for all. Currently the system is overloaded as most of the collections under the ansible-collections org run **all** their tests against every commit. If we switch to GHA, out test run will last up to a couple of hours...

Of course the implementation is discussable, i just can't imagine now how it'll look.
I see GHA implementation in c.mysql very convenient, the question is if it's possible to do something similar with AZP OR how to run targets in GHA separately - only for changed modules (and their unit/integration tests), not all the targets.

> 2. can i remove `tests/integration/targets/*/vars/*` for OSes that are not listed in the Azure pipeline?

Makes sense.

IMO ideally we use one distro, say, Ubuntu or CentOS and install and run all supported PG and connector versions on it.

hunleyd:
> > 1. does this approach make sense to you (installing all 5 versions in the test instance at once)?
> 
> I'm not sure i fully understand what the implementation plan is. Could you please explain it?

Well, based on what little I know about AZP it seemed like the solution was to simply run each of the existing tests 5 times (using `with_sequence`). So what I've done so far is install the PGDG repo onto the test image, and then I install PG 10 - 14 assigning each a listen port of 54nn where nn is the version of PG (so 5410 for PG10 5411 for PG11, etc). This gives us the five instances we want to test against. Then for each of the tests, in main.yml we'd do the same `with_sequence` loop and change the port we connect to on each iteration. Basically we'd tweak `tests/integration/targets/postgresql_copy/tasks/main.yml` (and the others) to run itself 5 times (one per port).

> IMO, ideally it should look like in [ansible-collections/community.mysql#370](https://github.com/ansible-collections/community.mysql/pull/370). If you open the test section, you'll see entries like `Integration (Python: 3.8, Ansible: stable-2.12, MySQL: mariadb_10.8.3, Connector: mysqlclient==2.0.1`.
> 
>     * So there we see separate tasks for each version of MySQL and the connector. I think it's very convenient to have them separate.

Yeah, this was what I had originally envisioned, but looking at the AZP bits it didn't seem setup to support that.

> 
>     * I'm not sure if this is achievable in AZP.
> 
>     * GHA is not an option if we don't find a way to run the tests using diff, i.e. only tests for changed modules must run, not for all. Currently the system is overloaded as most of the collections under the ansible-collections org run **all** their tests against every commit. If we switch to GHA, out test run will last up to a couple of hours...

Oh, eek. Of course, I was surprised that AZP ran for a **draft** PR. That seems wrong to me :)

> 
> 
> Of course the implementation is discussable, i just can't imagine now how it'll look. I see GHA implementation in c.mysql very convenient, the question is if it's possible to do something similar with AZP OR how to run targets in GHA separately - only for changed modules (and their unit/integration tests), not all the targets.
> 

Who would be the best AZP resource to go harass? :)

> > 2. can i remove `tests/integration/targets/*/vars/*` for OSes that are not listed in the Azure pipeline?
> 
> Makes sense.

OK

> 
> IMO ideally we use one distro, say, Ubuntu or CentOS and install and run all supported PG and connector versions on it.

I was actually thinking that myself and would prefer it.

Andersson007:
@hunleyd thanks for clarifying!

>Well, based on what little I know about AZP it seemed like the solution was to simply run each of the existing tests 5 times (using with_sequence). So what I've done so far is install the PGDG repo onto the test image, and then I install PG 10 - 14 assigning each a listen port of 54nn where nn is the version of PG (so 5410 for PG10 5411 for PG11, etc). This gives us the five instances we want to test against. Then for each of the tests, in main.yml we'd do the same with_sequence loop and change the port we connect to on each iteration. Basically we'd tweak tests/integration/targets/postgresql_copy/tasks/main.yml (and the others) to run itself 5 times (one per port).

With this approach, will it be easy (not only for us, folks who understand the process, but for newcomers) to see/find against which version of postgres/connector the tests fail? I think this is very important not to parse a whole log each time when our run fails.

hunleyd:
> With this approach, will it be easy (not only for us, folks who understand the process, but for newcomers) to see/find against which version of postgres/connector the tests fail? I think this is very important not to parse a whole log each time when our run fails.

define 'easy' :)

Will it say `CI (Docker 2.11 CentOS7 PG10)` in the checks? No. That would require having different images available to test with in AZP. no?

We can have the output of the tests say what PGnn is. Which is less than ideal, I agree.

Andersson007:
> Will it say CI (Docker 2.11 CentOS7 PG10) in the checks? No. That would require having different images available to test with in AZP. no?

I think to see a distro is not important but Ansible, PG and driver versions is important to see, otherwise it can be hard to troubleshoot

> We can have the output of the tests say what PGnn is. Which is less than ideal, I agree.

Yep, it's less than ideal.. it'll be a huge log for all the versions. On the other hand, it's also not ideal now but a bit better than having the huge log, now it's easier to find a version.

Andersson007:
But we can try what you're suggesting (imo) using:
- One distro, say, Ubuntu 20.04 (or any other which can run all the versions we need)
- We should have opportunity to add different connector versions later

and then see if it lives up to our demands

PR #324, opened 2022-07-27
added dsn field to allow additional connection parameters as a libpg string

Boosai:
##### SUMMARY

Added "dsn" field to allow connection parameters as a libpg string.

The postgres library "psycopg" used in this module allows to specify a
libpq connection string. It is named "dsn" there, so I took the same
name.

https://www.psycopg.org/docs/module.html#psycopg2.connect

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

postgres_*

Affects all modules that establish a database connection.

##### ADDITIONAL INFORMATION

The dsn parameter string is passed unmodified to the module.

Multiple connection parameters would be passed by separating each
parameter with a space, as decribed in the "psycopg" documentation.

For example:

There is a cluster setup "host1" and "host2". But we do not know which one
of the hosts is the primary (read-write access) and the secondary
(write-only access). We want to insert some data:

```
  - community.postgresql.postgresql_query:
      dsn: "target_session_attrs=read-write"
      login_host: "host1,host2"
      login_user: "test"
      login_password: "test1234"
      db: 'test'
      query: >
        insert into test (test) values (now())
```

We list both hosts in "login_host" and added
"target_session_attrs=read-write" as "dsn" to automatically select the
primary.

(Otherwise we would get "cannot execute INSERT in a read-only
transaction" if "host1" is currently the secondary.)

https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING

### DIFF
--- a/changelogs/fragments/324-postgresql_add_dsn_field.yml
+++ b/changelogs/fragments/324-postgresql_add_dsn_field.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - postgresql_* - added ``dsn`` field to allow additional connection parameters as a libpg string
--- a/plugins/doc_fragments/postgres.py
+++ b/plugins/doc_fragments/postgres.py
@@ -48,6 +48,10 @@ options:
       - If the file exists, the server's certificate will be verified to be signed by one of these authorities.
     type: str
     aliases: [ ssl_rootcert ]
+  dsn:
+    description:
+      - Connection parameters as a libpq connection string.
+    type: str
 notes:
 - The default authentication assumes that you are either logging in as or sudo'ing to the C(postgres) account on the host.
 - To avoid "Peer authentication failed for user postgres" error,
--- a/plugins/module_utils/postgres.py
+++ b/plugins/module_utils/postgres.py
@@ -44,6 +44,7 @@ def postgres_common_argument_spec():
         port=dict(type='int', default=5432, aliases=['login_port']),
         ssl_mode=dict(default='prefer', choices=['allow', 'disable', 'prefer', 'require', 'verify-ca', 'verify-full']),
         ca_cert=dict(aliases=['ssl_rootcert']),
+        dsn=dict(default=''),
     )
 
 
@@ -185,7 +186,8 @@ def get_conn_params(module, params_dict, warn_db_default=True):
         "login_password": "password",
         "port": "port",
         "ssl_mode": "sslmode",
-        "ca_cert": "sslrootcert"
+        "ca_cert": "sslrootcert",
+        "dsn": "dsn"
     }
 
     # Might be different in the modules:
--- a/plugins/modules/postgresql_query.py
+++ b/plugins/modules/postgresql_query.py
@@ -150,6 +150,15 @@ EXAMPLES = r'''
       id_val: 1
       story_val: test
 
+- name: Use dsn to add any connection parameters as a libpg string
+  community.postgresql.postgresql_query:
+    dsn: "target_session_attrs=read-write"
+    login_host: "host1,host2"
+    login_user: "test"
+    login_password: "test1234"
+    db: 'test'
+    query: 'insert into test (test) values (now())'
+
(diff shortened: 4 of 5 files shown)


hunleyd:
is this still a draft or ready for review @Boosai ?

hunleyd:
i _love_ the idea of this, and i like what you've done, but DSN is 'data source name' (the full connection string) and looks something like `parse_dsn("postgresql://someone@example.com/somedb?connect_timeout=10")` (taken from the psycopg2 docs; you can see the whole connect string is the DSN). This PR is adding the ability to add DSN _parameters_ right (e.g. connect_timeout, target_session_attrs)? So maybe `s/dns/dsn_params/` in your PR?

Ideally, we would also accept 'dsn' _instead of_ all the other params:
```
  - community.postgresql.postgresql_query:
      dsn: "test:test1234@example.com/testdb?target_session_attrs=read-write"
      query: >
        insert into test (test) values (now())
```
Would you want to tackle that too?

Boosai:
Yes I agree, the name `dsn` is somewhat confusing as associated with the full connection string.

While thinking about it, I tried a more generic approach.
What do you think if one is allowed to pass in any arbitrary parameters with a dict (connect_params)?

The advantage would be that one can specify a DSN with connection string, and/or add additional parameters.
The additional parameters can be added in a more readable way. They don't have to be assembled as one string.

Everything is passed unchanged to `psycopg2`.

The example above would be changed to:

```
  - community.postgresql.postgresql_query:
      connect_params:
        dsn: "postgresql://test:test1234@example.com/testdb?target_session_attrs=read-write&connect_timeout=10"
      query: "insert into test (test) values (now())"
```

And written in a more "classic" way:

```
  - community.postgresql.postgresql_query:
      connect_params:
        target_session_attrs: "read-write"
        connect_timeout: "10"
      login_host: "example.com"
      login_user: "test"
      login_password: "test1234"
      db: 'test'
      query: "insert into test (test) values (now())"
```

I started a new branch [plusserver:connect_params](https://github.com/plusserver/community.postgresql/tree/connect_params) that implements it.

If you like this approach more, I would make a new PR for it. (Or should I change this one?)

Boosai:
While testing I discovered one problem.

If you specify a DSN like `postgresql://test:test1234@example.com/testdb` the user is not `test` but `postgres`.

This is because `psycopg2` takes the `dsn` first and other kwargs
override fields in this. The ansible module defaults `login_user` to
`postgres` and therefor `test` is overwritten.
A similar thing would happen with other parameters that have default
values set (eg port).

I'm unsure how to deal with it.

- Parsing the dsn and making some (complicated?) logic?

- Removing defaults, make a breaking change?

- Describe and explain the behavior in the documentation?


A workaround for the current example would be:
```
  - community.postgresql.postgresql_query:
      connect_params:
        dsn: "postgresql://test:test1234@example.com/testdb?target_session_attrs=read-write&connect_timeout=10"
      login_user: ""
      query: "insert into test (test) values (now())"
```
Setting `login_user` to an empty string prevents to use its default.

hunleyd:
> I started a new branch plusserver:connect_params that implements it.
>
> If you like this approach more, I would make a new PR for it. (Or should I change this one?)

i like it! what do you think @Andersson007 ?

you can either create a new PR or update this one. Whatever works best w/ your personal workflow.

> While testing I discovered one problem.

Ugh, that's hairy. Maybe is 'dsn' is specified we unset the defaults? I don't wanna unset them permanently for everyone, that would be a pita for users. Ideas @Andersson007 ?

Boosai:
I started a new [PR#329](https://github.com/ansible-collections/community.postgresql/pull/329) for the "connect_params" approach.

Boosai:
Maybe this PR should be closed as PR#329 supersedes this one?!

hunleyd:
> Maybe this PR should be closed as PR#329 supersedes this one?!

makes sense to me

PR #329, opened 2022-08-05
added connect_params to allow any additional parameters that libpg understands

Boosai:
##### SUMMARY

Added "connect_params" parameter dict to allow any additional libpg connection parameters.

The postgres library "psycopg" used in this module allows to add
any parameters as kwargs.

The dict "connect_params" can be used to pass in any parameters as kwargs to "psycopg".

https://www.psycopg.org/docs/module.html#psycopg2.connect

This is a more generic approach of [PR#324](https://github.com/ansible-collections/community.postgresql/pull/324).

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
postgres_*

Affects all modules that establish a database connection.

##### ADDITIONAL INFORMATION

The key/values of the connect_params parameter dict is passed unmodified as kwargs to the module.

For example:

There is a cluster setup "host1" and "host2". But we do not know which one
of the hosts is the primary (read-write access) and the secondary
(write-only access). In addition we want to set the connection timeout. 
We want to insert some data:

```
  - community.postgresql.postgresql_query:
      connect_params:
        target_session_attrs: "read-write"
        connect_timeout: "10"
      login_host: "host1,host2"
      login_user: "test"
      login_password: "test1234"
      db: 'test'
      query: >
        insert into test (test) values (now())
```

We list both hosts in "login_host" and added
"target_session_attrs=read-write" as additional connection parameter to automatically select the
primary.

(Otherwise we would get "cannot execute INSERT in a read-only
transaction" if "host1" is currently the secondary.)

And we set the connection timeout to 10 seconds.

https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING

### DIFF
--- a/changelogs/fragments/329-postgresql_add_connect_params_field.yml
+++ b/changelogs/fragments/329-postgresql_add_connect_params_field.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  postgresql_* - Add ``connect_params`` parameter dict to allow any additional libpg connection parameters.
+- postgresql_* - Add ``connect_params`` parameter dict to allow any additional libpg connection parameters.
--- a/changelogs/fragments/329-postgresql_add_connect_params_field.yml
+++ b/changelogs/fragments/329-postgresql_add_connect_params_field.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  postgresql_* - Add ``connect_params`` parameter dict to allow any additional libpg connection parameters.
--- a/changelogs/fragments/329-postgresql_add_connect_params_field.yml
+++ b/changelogs/fragments/329-postgresql_add_connect_params_field.yml
@@ -1,2 +1,2 @@
 minor_changes:
-- postgresql_* - Add ``connect_params`` parameter dict to allow any additional libpg connection parameters.
+- postgresql_* - add the ``connect_params`` parameter dict to allow any additional ``libpg`` connection parameters (https://github.com/ansible-collections/community.postgresql/pull/329).
--- a/plugins/doc_fragments/postgres.py
+++ b/plugins/doc_fragments/postgres.py
@@ -51,6 +51,7 @@ options:
   connect_params:
     description:
       - Any additional parameters to be passed to libpg.
+      - These parameters take precedence.
     type: dict
     version_added: '2.3.0'
 notes:
--- a/plugins/doc_fragments/postgres.py
+++ b/plugins/doc_fragments/postgres.py
@@ -52,6 +52,7 @@ options:
     description:
       - Any additional parameters to be passed to libpg as kwargs.
     type: dict
+    version_added: '2.3.0'
 notes:
 - The default authentication assumes that you are either logging in as or sudo'ing to the C(postgres) account on the host.
 - To avoid "Peer authentication failed for user postgres" error,
--- a/plugins/doc_fragments/postgres.py
+++ b/plugins/doc_fragments/postgres.py
@@ -48,6 +48,10 @@ options:
       - If the file exists, the server's certificate will be verified to be signed by one of these authorities.
     type: str
     aliases: [ ssl_rootcert ]
+  connect_params:
+    description:
+      - Any additional parameters to be passed to libpg as kwargs.
+    type: dict
 notes:
 - The default authentication assumes that you are either logging in as or sudo'ing to the C(postgres) account on the host.
 - To avoid "Peer authentication failed for user postgres" error,
--- a/plugins/doc_fragments/postgres.py
+++ b/plugins/doc_fragments/postgres.py
@@ -50,7 +50,7 @@ options:
     aliases: [ ssl_rootcert ]
   connect_params:
     description:
-      - Any additional parameters to be passed to libpg as kwargs.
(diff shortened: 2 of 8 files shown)


plugins/doc_fragments/postgres.py
```diff
@@ -48,6 +48,10 @@ class ModuleDocFragment(object):
       - If the file exists, the server's certificate will be verified to be signed by one of these authorities.
     type: str
     aliases: [ ssl_rootcert ]
+  connect_params:
+    description:
+      - Any additional parameters to be passed to libpg as kwargs.
+    type: dict
```
hunleyd, line 54: @Andersson007, doesn't this need a 'version added'? 2.3.0?

Andersson007: @hunleyd yep, there should be `version_added: 2.3.0`, thanks for spotting

changelogs/fragments/329-postgresql_add_connect_params_field.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- postgresql_* - Add ``connect_params`` parameter dict to allow any additional libpg connection parameters.
```
Andersson007, line 2:
```suggestion
- postgresql_* - add the ``connect_params`` parameter dict to allow any additional ``libpg`` connection parameters (https://github.com/ansible-collections/community.postgresql/pull/329).
```
To make it consistent with other changelog entries

plugins/modules/postgresql_query.py
```diff
@@ -155,6 +155,17 @@
     db: test_db
     query: INSERT INTO test_table (id, story) VALUES (2, 'my_long_story')
 
+- name: Use connect_params to add any additional connection parameters that libpg supports
+  community.postgresql.postgresql_query:
+    connect_params:
+      target_session_attrs: "read-write"
+      connect_timeout: "10"
```
Andersson007, line 162:
```suggestion
      connect_timeout: 10
```

plugins/modules/postgresql_query.py
```diff
@@ -155,6 +155,17 @@
     db: test_db
     query: INSERT INTO test_table (id, story) VALUES (2, 'my_long_story')
 
+- name: Use connect_params to add any additional connection parameters that libpg supports
+  community.postgresql.postgresql_query:
+    connect_params:
+      target_session_attrs: "read-write"
```
Andersson007, line 161:
```suggestion
      target_session_attrs: read-write
```
I suggest eliminating quotes whenever possible as it's an extra thing (see the examples around)

Andersson007 (commented):
@Boosai hi, thanks for the new feature!

Could you also please add the new parameter to any task of any file of any target (say, postgresql_info/tasks/* or postgresql_query/tasks/*) in `tests/integration/targets`?

plugins/doc_fragments/postgres.py
Boosai: I've added the version info.

Boosai:
> Could you also please add the new parameter to any task of any file of any target (say, postgresql_info/tasks/* or postgresql_query/tasks/*) in `tests/integration/targets`?

@Andersson007, I added "connect_timeout=30" via "connect_param" to any tasks there.

Andersson007 (approved):
@Boosai LGTM thanks!

@hunleyd if you're OK with it, feel free to merge

plugins/doc_fragments/postgres.py
```diff
@@ -48,6 +48,11 @@ class ModuleDocFragment(object):
       - If the file exists, the server's certificate will be verified to be signed by one of these authorities.
     type: str
     aliases: [ ssl_rootcert ]
+  connect_params:
+    description:
+      - Any additional parameters to be passed to libpg as kwargs.
```
Andersson007, line 53:
```suggestion
      - Any additional parameters to be passed to libpg as kwargs.
```
Does `as kwargs` say anything to users? If not, let's remove it maybe?

Andersson007: Besides that LGTM (i hope there will not be conflicts with existing parameters like when `login_host` and connection_params' `host` will be both passed

Boosai: > Does `as kwargs` say anything to users? If not, let's remove it maybe?
Yes, that may be incomprehensible. I removed it.

Boosai: > i hope there will not be conflicts with existing parameters like when `login_host` and connection_params' `host` will be both passed

Parameters passed via `connect_params` override other parameters. 

In this case `login_host` is mapped to `host` param for libpg and `connect_params['host']` overrides the `host` param (for libpg). So `connect_params['host']` is taken as hostname.

Maybe we should leave a note in the description like this:?
`- These parameters take precedence over the others.`

Andersson007: sounds good, wfm, thanks!

hunleyd (approved):

jchancojr:
LGTM. Thx, Doug!

jchancojr (approved):
LGTM

PR #347, opened 2022-09-21
add 2.14 to test matrix

hunleyd:
Update test matrix as per [24](https://github.com/ansible-collections/news-for-maintainers/issues/24)

### DIFF
--- a/.azure-pipelines/azure-pipelines.yml
+++ b/.azure-pipelines/azure-pipelines.yml
@@ -23,7 +23,7 @@ schedules:
 
 variables:
   - name: checkoutPath
-    value: ansible_collections/community/postgresql
+    value: ansible_collections/ansible/posix
   - name: coverageBranches
     value: main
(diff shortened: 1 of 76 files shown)


hunleyd:
is that right @Andersson007 ?

Andersson007:
@hunleyd i think you grabbed a few extra commit that shouldn't be here. Also i think we shouldn't remove 2.11 as it's not EOL.
Could you please fix it, then will see ..how other collections will implement it:))

Andersson007:
i suggest leaving the blank lines there for readability

Andersson007:
@hunleyd we also need to add the same stuff for the code below the `Docker` and `Remote` sections and also add corresponding lines to the list at the bottom of the file

Andersson007 (approved):
LGTM

Andersson007:
@hunleyd thanks for the contribution!

PR #356, opened 2022-10-18
ci(azure pipeline): remove Ansible 2.11 from Azure pipeline

hunleyd:
remove Ansible 2.11 from Azure pipeline

### DIFF
--- a/tests/integration/targets/postgresql_subscription/defaults/main.yml
+++ b/tests/integration/targets/postgresql_subscription/defaults/main.yml
@@ -1,6 +1,6 @@
 pg_user: postgres
 db_default: postgres
-primary_port: 5433
+primary_port: 5431
 replica_port: 5434
 
 test_table1: acme1
--- a/tests/integration/targets/postgresql_subscription/defaults/main.yml
+++ b/tests/integration/targets/postgresql_subscription/defaults/main.yml
@@ -1,6 +1,5 @@
 pg_user: postgres
 db_default: postgres
-primary_port: 5431
 replica_port: 5434
 
 test_table1: acme1
--- a/.azure-pipelines/azure-pipelines.yml
+++ b/.azure-pipelines/azure-pipelines.yml
@@ -92,18 +92,6 @@ stages:
             - name: Units
               test: '2.12/units/1'
 
-  - stage: Ansible_2_11
-    displayName: Sanity & Units 2.11
-    dependsOn: []
-    jobs:
-      - template: templates/matrix.yml
-        parameters:
-          targets:
-            - name: Sanity
-              test: '2.11/sanity/1'
-            - name: Units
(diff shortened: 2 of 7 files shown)


Andersson007:
We should figure out what happened to the `postgresql_replication` module on Ubuntu 20.04.
The primary and replica can't start. Strange that it worked in the past.
@hunleyd do you have time to take a look? (no rush)

hunleyd:
yeah, i have no clue how to fix this @Andersson007

hunleyd:
for some reason, the ubuntu  boxes have both 14 and 15 installed, so the `find` is finding multiple `initdb` and multiple `pg_ctl` @Andersson007

hunleyd:
even after fixing the code to deal w/ finding multiple PG binaries, the cluster still won't start and I can't figure out why. I think I don't understand the flow of the tests well enough to debug further @Andersson007

Andersson007:
> even after fixing the code to deal w/ finding multiple PG binaries, the cluster still won't start and I can't figure out why. I think I don't understand the flow of the tests well enough to debug further @Andersson007

@hunleyd I'll take a look next week, np!

Andersson007:
@hunleyd hm, the tests with `--docker ubuntu2004` work (and pass) locally...

tests/integration/targets/postgresql_subscription/defaults/main.yml
```diff
@@ -1,6 +1,6 @@
 pg_user: postgres
 db_default: postgres
-primary_port: 5433
+primary_port: 5431
```
Andersson007, line 3:
```suggestion
primary_port: 5431
```
if it works, we could try to remove it from here as it's already defined in `tests/integration/targets/setup_postgresql_replication/defaults/main.yml` (i'm not 100% sure but it should work w/o having it here as well as in the same file of postgresql_info target)

Andersson007: no need to remove it now but we could try at the end, after CI is green

hunleyd:
how does `postgresql-14` work in your pr but not mine @Andersson007 ?

Andersson007:
> how does `postgresql-14` work in your pr but not mine @Andersson007 ?

@hunleyd in the AZP log, it shows `"No package matching 'apt-utils - postgresql-14' is available"`, i.e. that the apt module can't install  the package with the name `apt-utils - postgresql-14`, which is expected, I think:)
I can see a tiny difference in the `pg_package_list` lists from our files:) PTAL

tests/integration/targets/postgresql_subscription/defaults/main.yml
hunleyd: ok, trying this now

jchancojr (approved):
Interesting find! lgtm

PR #378, opened 2022-12-01
postgresql_pg_hba: fix false "changed" when called with "overwrite: true"

### LINKED ISSUE #376: postgresql_pg_hba is not idempotent when used with overwrite and rules
Consider the following role files

default/main.yml

```yaml
pgsql_dir: /var/lib/pgsql
pgsql_password_encryption: 'scram-sha-256'
pgsql_pg_hba_overwrite: true
pgsql_pg_hba_rules: [] # add custom rules here
pgsql_pg_hba_default_rules: # do not change these!
  - comment: Postgres Service User Socket Passwordless
    users: postgres
    databases: all
    contype: local
    method: peer

  - comment: Local Unix Socket Hashed Passwords
    users: all
    databases: all
    contype: local
    method: "{{ pgsql_password_encryption }}"

  - comment: Local IPv4 Hashed Passwords
    users: all
    databases: all
    contype: host
    address: 127.0.0.1/32
    method: "{{ pgsql_password_encryption }}"

  - comment: Local IPv6 Hashed Passwords
    users: all
    databases: all
    contype: host
    address: ::1/128
    method: "{{ pgsql_password_encryption }}"
```

tasks/main.yml

```yaml
- name: configure pg_hba
  community.postgresql.postgresql_pg_hba:
    dest: "{{ pgsql_dir }}/data/pg_hba.conf"
    keep_comments_at_rules: true
    overwrite: "{{ pgsql_pg_hba_overwrite }}"
    rules: "{{ pgsql_pg_hba_default_rules + pgsql_pg_hba_rules }}"
    rules_behavior: combine
  become: yes
  become_user: postgres
  notify: reload postgresql
```

When overwrite is set to `true` the task will always show as changed even if the file is identical to what it was before.  This means that the `reload postgresql` handler will run every time regardless of actual changes.

betanummeric:
#### postgresql_pg_hba bugfix

fixes #376

- When parsing the `pg_hba.conf`, trim whitespace between the rule and comment in the same line, to ensure equality tests work.
- Determine whether something changed by checking if read and to-be-written rules differ. Previously the check was slightly different: When `overwrite` was enabled, the rule list was emptied and adding rules again counted as change.

### DIFF
--- a/changelogs/fragments/378-postgresql_pg_hba_fix_change_detection.yml
+++ b/changelogs/fragments/378-postgresql_pg_hba_fix_change_detection.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - postgresql_pg_hba - fix ``changed`` return value for when ``overwrite`` is enabled (https://github.com/ansible-collections/community.postgresql/pull/378).
--- a/plugins/modules/postgresql_pg_hba.py
+++ b/plugins/modules/postgresql_pg_hba.py
@@ -337,6 +337,7 @@ class PgHba(object):
         # users, this might be totally off, but at least it is some info...
         self.users = set(['postgres'])
 
+        self.preexisting_rules = None
         self.read()
 
     def clear_rules(self):
@@ -381,6 +382,7 @@ class PgHba(object):
                         except PgHbaRuleError:
                             pass
             self.unchanged()
+            self.preexisting_rules = dict(self.rules)
         except IOError:
             pass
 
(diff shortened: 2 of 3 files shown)


hunleyd (approved):

Andersson007 (commented):
@betanummeric is it possible to cover the case with integration tests? If it's tricky, no problem

betanummeric:
It should be fairly easy to add a test, I can do it tomorrow.

Andersson007 (approved):
LGTM

@nlvw thanks for reporting the issue!
@hunleyd thanks for reviewing!
@betanummeric thanks for fixing this! I think you can merge the PR yourself as there are approvals (choose "Squash and Merge", not the other options)

PR #357, opened 2022-10-20
postgresql_set: get rid of extra quoting when passing list values

Andersson007:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.postgresql/issues/78

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
postgresql_set

Issue #78, opened 2021-03-31
community.postgresql.postgresql_set incorrectly quotes strings containing commas


nergdron:
just tested the latest release (`1.2.0`) since it was only released 15h ago, devel doesn't have any notable changes i can see which would affect this bug.

##### SUMMARY
when using `postgresql_set` with a string value that contains a comma, for some reason it puts extra doublequotes (`"`) around the string, breaking it when it's included in `postgresql.auto.conf`:

```
password_encryption = 'scram-sha-256'
shared_preload_libraries = '"pg_stat_statements,repmgr"'
```

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
community.postgresql.postgresql_set

##### ANSIBLE VERSION
```paste below
ansible 2.10.6
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/tessa/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/tessa/.local/lib/python3.9/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.9.2 (default, Feb 20 2021, 18:40:11) [GCC 10.2.0]
```

##### CONFIGURATION
```
```

##### OS / ENVIRONMENT
EndeavourOS (Arch variant) running Ansible from system `ansible` package, version 3.1.0.

##### STEPS TO REPRODUCE
in role `defaults/main.yml`:
```yaml
postgres:
  config:
    password_encryption: scram-sha-256
    shared_preload_libraries: pg_stat_statements,repmgr
```

and in role `tasks/main.yml`
```yaml
- name: configure postgresql
  community.postgresql.postgresql_set:
    name: "{{ item.key }}"
    value: "{{ item.value }}"
  become_user: postgres
  loop: "{{ postgres.config | dict2items }}"
  notify: restart postgres
```

##### EXPECTED RESULTS
keys with values that contain commas get correctly rendered out to postgresql and entered into the `postgresql.auto.conf` without extraneous quotes.

##### ACTUAL RESULTS
for some reason module renders values with commas with extra double quotes around them, breaking postgres config. I've verified that if I change the above config line to the following, it works correctly and doesn't inject extra quotes:
```yaml
    shared_preload_libraries: repmgr
```

interestingly, ansible doesn't show the extra quotes in the command output, they only show up in the destination file itself:
```paste below
TASK [postgres : configure postgresql] ********************************************
[...]

PR #400, opened 2023-01-21
Fix postgresql_set module if single-value param contains comma in value

RealGreenDragon:
##### SUMMARY

The PR #357 fix the multi-value parameters management (issue #78), but the new check assumes as multi-value each parameter with a comma in the value, that is incorrect as there are single-value parameters with comma in value.

If a single-value parameter is treated as a multi-value, the 'param_set' function builds an ALTER SYSTEM SET command with multiple comma-separated values, that fails with the message:
```
ERROR:  SET log_line_prefix takes only one argument
```

Single-value parameters with possible comma in value are  "*_command" and "*_prefix" that for exmple in v12 are:
- ssl_passphrase_command
- archive_command
- restore_command
- archive_cleanup_command
- recovery_end_command
- log_line_prefix

The most critical (that I add in tests) is 'log_line_prefix' because often it contains comma and a space at the end.

I simply fix the check to evaluate as single-value all parameters that ends with '_command' and '_prefix'.

If the fix is accepted, It can be back-ported also in stable branch.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

postgresql_set

### DIFF
--- a/changelogs/fragments/0-postgresql_set_avoid_wrong_values.yml
+++ b/changelogs/fragments/0-postgresql_set_avoid_wrong_values.yml
@@ -1,2 +1,2 @@
 bugfixes:
-- postgresql_set - avoid wrong values for single-value parameters containing commas (https://github.com/ansible-collections/community.postgresql/pull/357).
+- postgresql_set - avoid wrong values for single-value parameters containing commas (https://github.com/ansible-collections/community.postgresql/pull/400).
--- a/changelogs/fragments/0-postgresql_set_avoid_wrong_values.yml
+++ b/changelogs/fragments/0-postgresql_set_avoid_wrong_values.yml
@@ -1,0 +1,2 @@
+bugfixes:
+- postgresql_set - avoid wrong values for single-value parameters containing commas.
--- a/changelogs/fragments/0-postgresql_set_avoid_wrong_values.yml
+++ b/changelogs/fragments/0-postgresql_set_avoid_wrong_values.yml
@@ -1,2 +1,2 @@
 bugfixes:
-- postgresql_set - avoid wrong values for single-value parameters containing commas.
+- postgresql_set - avoid wrong values for single-value parameters containing commas (https://github.com/ansible-collections/community.postgresql/pull/357).
--- a/plugins/modules/postgresql_set.py
+++ b/plugins/modules/postgresql_set.py
@@ -315,7 +315,7 @@ def param_set(cursor, module, name, value, context):
         if str(value).lower() == 'default':
             query = "ALTER SYSTEM SET %s = DEFAULT" % name
         else:
-            if isinstance(value, str) and ',' in value and not name.endswith('_command') and not name.endswith('_prefix'):
+            if isinstance(value, str) and ',' in value and not name.endswith(('_command', '_prefix')):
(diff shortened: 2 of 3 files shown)


hunleyd:
```
set pgaudit.log = 'write, ddl';
set pgaudit.log_relation = on;
```
would still be handled incorrectly, yes?

RealGreenDragon:
Hi hunleyd,

the parameters pgaudit.log and pgaudit.log_relation are not affected by my PR.
pgaudit.log_relation does not contains commas, so always go to the standard management.
pgaudit.log contains commas but is multi-value, so it is correctly not not affected by my PR (i.e. still go to specific management).

Small example with python3 and pg 12.13 (from postgres user):
```
$ python3
>>> value='role, ddl'
>>> ','.join(["'" + elem.strip() + "'" for elem in value.split(',')])
"'role','ddl'"
>>>

$ psql
postgres=# alter system set pgaudit.log = 'role','ddl';
ALTER SYSTEM

$ grep "pgaudit.log =" 12/data/postgresql.auto.conf
pgaudit.log = 'role, ddl'
```

hunleyd (approved):

changelogs/fragments/0-postgresql_set_avoid_wrong_values.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+- postgresql_set - avoid wrong values for single-value parameters containing commas.
```
Andersson007, line 2:
```suggestion
- postgresql_set - avoid wrong values for single-value parameters containing commas (https://github.com/ansible-collections/community.postgresql/pull/357).
```

plugins/modules/postgresql_set.py
```diff
@@ -315,7 +315,7 @@ def param_set(cursor, module, name, value, context):
         if str(value).lower() == 'default':
             query = "ALTER SYSTEM SET %s = DEFAULT" % name
         else:
-            if isinstance(value, str) and ',' in value:
+            if isinstance(value, str) and ',' in value and not name.endswith('_command') and not name.endswith('_prefix'):
```
Andersson007, line 318:
```suggestion
            if isinstance(value, str) and ',' in value and not name.endswith(('_command', '_prefix')):
```
would this achieve the same logic?

Andersson007 (commented):
@RealGreenDragon hi, welcome and thanks for the PR!

A couple of minor things from me.

plugins/modules/postgresql_set.py
RealGreenDragon: Yes, same logic but nicer code, thanks !!

RealGreenDragon:
@Andersson007 Suggestions applied, thank you very much!

changelogs/fragments/0-postgresql_set_avoid_wrong_values.yml
RealGreenDragon: This is PR 400 not 357, I correct it.

Andersson007 (approved):

jchancojr (approved):

hunleyd:
thx @RealGreenDragon !

PR #412, opened 2023-02-10
postgresql_query, postgresql_script: let psycopg2 do the list -> pg array conversion

### LINKED ISSUE #411: postgresql_query, postgresql_script: array argument conversion
The postgresql_query and postgresql_script modules use this code to transform a list in the arguments to an postgres array string: https://github.com/ansible-collections/community.postgresql/blob/2.3.2/plugins/module_utils/postgres.py#L387-L422

## quoting issue

The current implementation doesn't work for a list of strings, because the single quotes from python become part of the array value:

```yaml
- hosts: all
  gather_facts: no
  remote_user: postgres
  tasks:
    - name: create one rule to clear
      community.postgresql.postgresql_query:
        login_unix_socket: /tmp
        query: |
          select unnest(%(input_list)s::text[])
        named_args:
          input_list:
            - a
            - b 
      register: query_result
    - name: show result
      debug:
        msg: "{{ query_result }}"
```

results in

```
...
        "query": "select unnest('{''a'', ''b''}'::text[])\n",
        "query_list": [
            "select unnest(%(input_list)s::text[])\n"
        ],
        "query_result": [
            {
                "unnest": "'a'"
            },
            {
                "unnest": "'b'"
            }
        ],
        "rowcount": 2,
...
```

while I would expect something like
```
...
        "query": "select unnest('{"a","b"}'::text[])\n",
        "query_list": [
            "select unnest(%(input_list)s::text[])\n"
        ],
        "query_result": [
            {
                "unnest": "a"
            },
            {
                "unnest": "b"
            }
        ],
        "rowcount": 2,
...
```

because single and double quotes are not the same in postgres:
```
postgres=# select '{''a'',b}'::text[] = '{a,b}'::text[];
 ?column? 
----------
 f
(1 row)

Time: 0.538 ms
postgres=# select '{"a",b}'::text[] = '{a,b}'::text[];
 ?column? 
----------
 t
(1 row)

Time: 0.782 ms
```

## ways to fix

psycopg2 knows how to convert python lists into postgres arrays: https://www.psycopg.org/docs/usage.html#adapt-list
I think we could just skip the conversion done in ansible and leave it to psycopg2.

If there is a reason to keep the conversion in ansible, I propose to change it from
```python3
def list_to_pg_array(elem):
    """Convert the passed list to PostgreSQL array
    represented as a string.
    Args:
        elem (list): List that needs to be converted.
    Returns:
        elem (str): String representation of PostgreSQL array.
    """
    elem = str(elem).strip('[]')
    elem = '{' + elem + '}'
    return elem
```

to

```python3
def list_to_pg_array(py_list):
  return '{' + ','.join([ '"' + str(item).replace('"', '\\"') + '"' for item in py_list ]) +  '}'
```

betanummeric:
fixes #411

This PR removes the logic which converts a query argument from a python list to a postgres array, because psycopg2 can do this better.

This changes the module arguments to SQL translation, so I'm not sure if this should be categorized as just a bugfix or a major change.

### DIFF
--- a/plugins/module_utils/postgres.py
+++ b/plugins/module_utils/postgres.py
@@ -396,44 +396,6 @@ def set_search_path(cursor, search_path):
     cursor.execute('SET search_path TO %s' % search_path)
 
 
-def convert_elements_to_pg_arrays(obj):
-    """Convert list elements of the passed object
-    to PostgreSQL arrays represented as strings.
-
-    Args:
-        obj (dict or list): Object whose elements need to be converted.
-
-    Returns:
-        obj (dict or list): Object with converted elements.
-    """
-    if isinstance(obj, dict):
-        for (key, elem) in iteritems(obj):
-            if isinstance(elem, list):
-                obj[key] = list_to_pg_array(elem)
-
-    elif isinstance(obj, list):
-        for i, elem in enumerate(obj):
-            if isinstance(elem, list):
-                obj[i] = list_to_pg_array(elem)
-
-    return obj
-
-
-def list_to_pg_array(elem):
-    """Convert the passed list to PostgreSQL array
-    represented as a string.
-
-    Args:
-        elem (list): List that needs to be converted.
-
-    Returns:
-        elem (str): String representation of PostgreSQL array.
-    """
-    elem = str(elem).strip('[]')
-    elem = '{' + elem + '}'
-    return elem
-
-
 def convert_to_supported(val):
     """Convert unsupported type to appropriate.
     Args:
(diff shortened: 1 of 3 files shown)


Andersson007:
@betanummeric if it changes the **existing** user interface anyhow (like values, choices, return values, etc.) / can lead to playbooks that worked won't work anymore after upgrade to a version containing the proposed change  - it'd be a major change

betanummeric:
uhh, this PR is quite old already. But I think it is still justified.
The change size categorization mostly depends on whether we call the existing behavior a bug or a feature. It it like that for quite a while now, so it's probably a feature now -> major change. :/
I guess the PR is in draft only because it needs some testing.

PR #422, opened 2023-02-27
fix: Make pretty_to_bytes return IP addresses as is, instead of throwing ValueError

sverrehu:
##### SUMMARY
During dry-run, `postgresql_set` may be confused and throw `ValueError` for certain values:

```yaml
community.postgresql.postgresql_set:
  name: "listen_addresses"
  value: "0.0.0.0"
```

gives

```text
ValueError: could not convert string to float: '0.0.0.0'
```

Errors like this [have been fixed before](https://github.com/ansible-collections/community.postgresql/pull/52), but that did not cover IP addresses. This PR will handle the `ValueError` that may result from `float(pretty_val)`, and just return `pretty_val` as is. 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/postgresql_set.py

### DIFF
--- a/plugins/modules/postgresql_set.py
+++ b/plugins/modules/postgresql_set.py
@@ -269,7 +269,11 @@ def pretty_to_bytes(pretty_val):
             pretty_val = int(pretty_val)
 
         except ValueError:
-            pretty_val = float(pretty_val)
+            try:
+                pretty_val = float(pretty_val)
+
+            except ValueError:
+                return pretty_val
 
         return pretty_val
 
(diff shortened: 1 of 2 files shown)


hunleyd (approved):

hunleyd:
needs a changelog fragment, right @Andersson007 ?

jchancojr (approved):
lgtm pending fragment

sverrehu:
> lgtm pending fragment

Thank you! Is that something I'm supposed to provide?

Andersson007:
@sverrehu hi, thanks for the fix!
Could you please add the fragment using https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment, so that our users will know about the fix

sverrehu:
Like this?

hunleyd (approved):

hunleyd:
Perfect @sverrehu ! TYVM

Andersson007:
@sverrehu thanks for the contribution! @hunleyd @jchancojr thanks for reviewing, merging and backporting!

PR #502, opened 2023-06-09
Privs public role

RealGreenDragon:
##### SUMMARY

I checked that sometime postgresql_privs module is not idempotent (always return changed=True).

I found the cause in manipulate_privs function:
```
if roles == 'PUBLIC':
    return True
```

Seems who inserted this "strange if" thought that there is no changes in ACLs if the grantee is PUBLIC, but it is completely wrong. A grant to PUBLIC results in an ACL entry with name empty, so the check in manipulate_privs function manage to find the change.

Example:
```
test=# SELECT * FROM pg_catalog.pg_parameter_acl;
 oid | parname | paracl
-----+---------+--------
(0 rows)

test=# GRANT ALTER SYSTEM ON PARAMETER logging_collector TO PUBLIC;
GRANT

test=# SELECT * FROM pg_catalog.pg_parameter_acl;
  oid  |      parname      |         paracl
-------+-------------------+------------------------
 16389 | logging_collector | {test=sA/test,=A/test}
(1 row)

test=# CREATE USER example;
CREATE ROLE

test=# GRANT ALTER SYSTEM ON PARAMETER logging_collector TO example;
GRANT

test=# SELECT * FROM pg_catalog.pg_parameter_acl;
  oid  |      parname      |                paracl
-------+-------------------+---------------------------------------
 16389 | logging_collector | {test=sA/test,=A/test,example=A/test}
(1 row)
```

But also if who inserted this "strange if" thought that there is no changes in ACLs if the grantee is PUBLIC (and repeat, it is completely wrong), the check does not cover the case of multiple roles.

Without this "strange if" (sure an old wrong)  all is ok and works as expected.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

postgresql_privs

##### ADDITIONAL INFORMATION

An harmful if removed.


RealGreenDragon:
Obviously the string "X.X.X" in comments added must be replaced with the collection release that will include this PR.

RealGreenDragon:
Ready for review

hunleyd (approved):
i defer to @Andersson007 but i think you can use 3.0.1 for your x.x.x

Andersson007:
@RealGreenDragon @hunleyd 
folks, i'm not a user, so i cannot judge these changes from user prespective.
i see you put this under `major_changes`, does it break the current interface?
Put another way, will users get (or can potentially get) their playbooks broken after they update ansible or the collection after this change is released?

plugins/modules/postgresql_privs.py
```diff
@@ -77,8 +77,11 @@
   roles:
     description:
     - Comma separated list of role (user/group) names to set permissions for.
-    - The special value C(PUBLIC) can be provided instead to set permissions
-      for the implicitly defined PUBLIC group.
+    - Each role must be in lowercase as internally PostgreSQL store them in lowercase only.
```
betanummeric, line 80: PostgreSQL stores roles case-sensitive. You can use any random 63 bytes as a role name, as long as it's quoted properly, see https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS.
If an identifier is not quoted, it is converted to lowercase.
Quoting also allows to have roles with confusing names:
```SQL
postgres=# select version();
                                              version                                              
---------------------------------------------------------------------------------------------------
 PostgreSQL 15.1 on x86_64-pc-linux-gnu, compiled by gcc (Debian 10.2.1-6) 10.2.1 20210110, 64-bit
(1 row)
postgres=# create role public;
ERROR:  role name "public" is reserved
LINE 1: create role public;
                    ^
Time: 1.106 ms
postgres=# create role "Public";
CREATE ROLE
Time: 4.282 ms
```

We should check if quoted role names are supported by this module. It's not a limitation of PostgreSQL.

changelogs/fragments/502_postgresql_privs.yml
```diff
@@ -0,0 +1,2 @@
+major_changes:
+- "postgresql_privs - Improved roles management (https://github.com/ansible-collections/community.postgresql/pull/502)."
```
betanummeric, line 2: Please elaborate. The changelog should describe all changes in the behavior and interface of the module. Only refactoring can be omitted here.

plugins/modules/postgresql_privs.py
```diff
@@ -1120,17 +1115,25 @@ def main():
...
 
         # roles
-        if p.roles.upper() == 'PUBLIC':
-            roles = 'PUBLIC'
-        else:
-            roles = p.roles.split(',')
-
-            if len(roles) == 1 and not role_exists(module, conn.cursor, roles[0]):
+        roles = []
+        roles_raw = p.roles.split(',')
```
betanummeric, line 1119: Role names containing a `,` are supported by PostgreSQL. The safest way to deal with that would probably be to extend the `roles` module argument to also take a list of strings, something like:
```python3
if isinstance(p.roles, list):
  roles_raw = list(p.roles)
else:
  roles_raw = p.roles.split(',')
```

plugins/modules/postgresql_privs.py
RealGreenDragon: You are right, I forgot to fix docs (the code is already ok).
Done in e408a01a853eb5600b859530e59f802da75c72ac.

RealGreenDragon:
I fixed what requested, thanks for all advices.
I'm unable to implement list support for parameter "roles" as [here](https://github.com/ansible-collections/community.postgresql/pull/502/files#r1228585734).
I'm waiting to know collection version to replace at X.X.X.

plugins/modules/postgresql_privs.py
```diff
@@ -77,8 +77,9 @@
   roles:
     description:
     - Comma separated list of role (user/group) names to set permissions for.
-    - The special value C(PUBLIC) can be provided instead to set permissions
-      for the implicitly defined PUBLIC group.
+    - Roles C(PUBLIC), C(CURRENT_ROLE), C(CURRENT_USER), C(SESSION_USER) are implicitly defined in PostgreSQL.
+    - C(CURRENT_USER) and C(SESSION_USER) implicit roles are supported since collection version X.X.X and PostgreSQL 9.5.
+    - C(CURRENT_ROLE) implicit role is supported since collection version X.X.X and PostgreSQL 14.
```
Andersson007, line 82:
```suggestion
    - C(CURRENT_USER) and C(SESSION_USER) implicit roles are supported since collection version 3.1.0 and PostgreSQL 9.5.
    - C(CURRENT_ROLE) implicit role is supported since collection version 3.1.0 and PostgreSQL 14.
```

plugins/modules/postgresql_privs.py
```diff
@@ -1056,7 +1049,7 @@ def main():
     except TypeError as e:
         if 'sslrootcert' in e.args[0]:
             module.fail_json(msg='Postgresql server must be at least version 8.4 to support sslrootcert')
-        module.fail_json(msg="unable to connect to database: %s" % to_native(e), exception=traceback.format_exc())
+        module.fail_json(msg="Unable to connect to database: %s" % to_native(e), exception=traceback.format_exc())
```
Andersson007, line 1052:
```suggestion
        module.fail_json(msg="unable to connect to database: %s" % to_native(e), exception=traceback.format_exc())
```
let's not mix unrelated changes in one PR

If there are any other changes that can be done in separate PRs, please remove them from here and open other PRs

Andersson007 (changes requested):
@RealGreenDragon thanks for pushing forward

So do we expect these changes keep backwards compatibility? i.e. users playbooks will not break after upgrading the collection?

RealGreenDragon:
@Andersson007 I confirm this PR is a minor change. I only added other implicit roles (before not supported) and fixed idempotence when roles=PUBLIC (now it reports changed only when really changed). In addiction I refactored a bit 'roles' parameter management, but without breacking changes.

betanummeric (changes requested):
PostgreSQL allows almost all role names when quoted:

```
test=# drop role "SESSION_USER";
DROP ROLE
Time: 2.079 ms
test=# create role "PUBLIC";
CREATE ROLE
Time: 1.997 ms
test=# create role "CURRENT_USER";
CREATE ROLE
Time: 1.829 ms
test=# create role "CURRENT_ROLE";
CREATE ROLE
Time: 1.719 ms
test=# create role "SESSION_ROLE";
CREATE ROLE
Time: 1.854 ms
test=# create role "public";
ERROR:  role name "public" is reserved
LINE 1: create role "public";
                    ^
Time: 1.049 ms
test=# create role "current_user";
CREATE ROLE
Time: 2.511 ms
```
So if someone was using such an confusing role name, this PR would break it. But I think we can treat it as a minor change.

plugins/modules/postgresql_privs.py
betanummeric: @RealGreenDragon 
We could still make this work by making the ansible module argument `type='raw'`, which disables type checking by ansible for that. Then we should be able to check the type "on our own" with something like to code example from my comment above.

I'm unresolving this in case you want to implement it in this PR. Doing it in some other PR would also be fine for me.

RealGreenDragon:
All advices implemented.
RHEL tests still failed, seems py3.9 cannot be installed.

Andersson007 (approved):

Andersson007:
I approved the PR and it's about general stuff.
@hunleyd @betanummeric do you think it's ready for merge? No pressure: considering importance of the module, feel free to add more comments of course

jchancojr (approved):

Andersson007:
cc @hunleyd @betanummeric

hunleyd:
@Andersson007 assuming the open conversation from @betanummeric is addressed, i'm ok with merging it

hunleyd:
merged as per @Andersson007's comment on matrix

betanummeric:
Yes, this looks ready to be merged.

Andersson007:
@RealGreenDragon thanks for the great contribution!
@hunleyd @betanummeric @jchancojr thanks for reviewing!

PR #517, opened 2023-06-28
Psycopg 3 support

elpavel:
##### SUMMARY
Added support for Psycopg 3

Fixes #499

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
All modules

##### ADDITIONAL INFORMATION
Work in progress, any feedback welcome

Issue #499, opened 2023-06-09
Add Psycopg 3 support


Andersson007:
##### SUMMARY
- Related PR https://github.com/ansible-collections/community.postgresql/pull/220 that can be used when implementing actual support (the obstacle was fixed on the driver's stuff)
- Another big related issue to solve it adding testing against psycopg 3, i.e. to have testing against 2 and 3

elpavel:
It looks like the Linux distributions don't include psycopg 3 yet (except debian/bookworm). While the distros are catching up and psycopg 3 becoming available, it may be enough to have just one test target with both versions of psycopg available. The existing modules will still work, and can be slowly updated one by one. Assuming the updated modules will try to import version 3 first. Broader testing of psycopg 3 can be added later when packages become widely available.
So an easy way to enable testing against psycopg 3 could be
- add a new target (let say ubuntu2204)
- keep psycopg2 for this target
- add psycopg 3 on this target via Pip

Andersson007:
@elpavel yep, new target with ubuntu2204 sounds like a great idea and we could specify in the `setup_postgresql_db` target a variable to install `psycopg3` when there's `ubuntu2204` (though i didn't test this).

When i took the first try in https://github.com/ansible-collections/community.postgresql/pull/220, i stumbled upon one incompatibility in `psycopg3` which would require a lot of code re-writing.
I discussed that with their devs and their added a toggle (a variable afaik) which allows to say the driver to keep the compatibility on the initialization stage. So it should be quite easy to add the support now (see my PR).

Would you like to pick it up? Anyway, thanks for your participation!

Andersson007:
@elpavel IIRC yes,
Thanks a lot!

elpavel:
@Andersson007 @hunleyd should we consider supporting all (old) releases of psycopg2 ?

2.5 was released in 2013
2.8 was released in 2019

The two above releases added functionality which is also present in Psycopg 3, and using that functionality would simplify the code as the same would work for Psycopg 2 and 3.

Examples:
[psycopg2.Error.diag](https://www.psycopg.org/docs/module.html#psycopg2.Error.diag) added in 2.5
[psycopg2.errors](https://www.psycopg.org/docs/errors.html#module-psycopg2.errors) added in 2.8

hunleyd:
> @Andersson007 @hunleyd should we consider supporting all (old) releases of psycopg2 ?
> 
> 2.5 was released in 2013 2.8 was released in 2019
> 
[...]

PR #523, opened 2023-07-11
Psycopg dictionary results

elpavel:
##### SUMMARY
Modifying all psycopg results to use dictionary keys, and casting all ACL type data as varchar. 

This will ensure the same code will work with Psycopg 3, helping with #499. 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Many modules

##### ADDITIONAL INFORMATION
These changes were tested with Psycopg 3 in #517.

### DIFF
--- a/plugins/modules/postgresql_info.py
+++ b/plugins/modules/postgresql_info.py
@@ -691,7 +691,7 @@ class PgClusterInfo(object):
                              "WHERE table_schema = 'pg_catalog' "
                              "AND table_name = 'pg_subscription'")
         columns_result = self.__exec_sql(columns_sub_table)
-        columns = ", ".join(["s.%s" % column[0] for column in columns_result])
+        columns = ", ".join(["s.%s" % column["column_name"] for column in columns_result])
 
         query = ("SELECT %s, r.rolname AS ownername, d.datname AS dbname "
                  "FROM pg_catalog.pg_subscription s "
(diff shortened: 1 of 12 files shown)


Andersson007 (approved):
@elpavel i like the gradual approach you use.
If we're sure it's backwards compatible and will not touch user workflows, this refactoring LGTM
thanks

plugins/modules/postgresql_info.py
```diff
@@ -731,22 +731,22 @@ def get_tablespaces(self):
                               "AND column_name = 'spcoptions'")
 
         if not opt:
-            query = ("SELECT s.spcname, pg_catalog.pg_get_userbyid(s.spcowner) as rolname, s.spcacl "
+            query = ("SELECT s.spcname, pg_catalog.pg_get_userbyid(s.spcowner) as rolname, s.spcacl::varchar "
```
hunleyd, line 734: why varchar and not text? the field is of type aclitem and in pg varchar is just text w/ input limits

hunleyd: it probably doesn't make a difference, i'm just curious.

hunleyd (changes requested):

plugins/modules/postgresql_info.py
elpavel: I wasn't sure which one to use, whether text or varchar, so picked varchar by habit. 
Happy to change it, shouldn't make a difference.

Andersson007 (approved):

PR #533, opened 2023-07-14
Psycopg autocommit handling

elpavel:
##### SUMMARY
Improved autocommit handling, to work with Psycopg 3 when we add the support.
It will help with #499

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
postgresql_tablespace

##### ADDITIONAL INFORMATION

The following doesn't work in Psycopg 3
```
connection.set_session(autocommit=True)
```

it is being replaced with
```
connection.autocommit = True
```
both methods were introduced in psycopg 2.4.2 https://www.psycopg.org/docs/connection.html#connection.autocommit

The code has been tested with Psycopg 3 in #517

### DIFF
--- a/changelogs/fragments/533-postgresql_tablespace.yml
+++ b/changelogs/fragments/533-postgresql_tablespace.yml
@@ -1,2 +1,0 @@
-minor_changes:
-- "postgresql_tablespace - improved autocommit handling (https://github.com/ansible-collections/community.postgresql/pull/533)."
--- a/changelogs/fragments/533-postgresql_tablespace.yml
+++ b/changelogs/fragments/533-postgresql_tablespace.yml
@@ -1,2 +1,2 @@
-bugfixes:
+minor_changes:
 - "postgresql_tablespace - improved autocommit handling (https://github.com/ansible-collections/community.postgresql/pull/533)."
--- a/changelogs/fragments/533-postgresql_tablespace.yml
+++ b/changelogs/fragments/533-postgresql_tablespace.yml
@@ -1,0 +1,2 @@
+bugfixes:
+- "postgresql_tablespace - improved autocommit handling (https://github.com/ansible-collections/community.postgresql/pull/533)."
--- a/plugins/modules/postgresql_tablespace.py
+++ b/plugins/modules/postgresql_tablespace.py
@@ -179,7 +179,6 @@ state:
 '''
 
 try:
-    from psycopg2 import __version__ as PSYCOPG2_VERSION
     from psycopg2.extensions import ISOLATION_LEVEL_AUTOCOMMIT as AUTOCOMMIT
     from psycopg2.extensions import ISOLATION_LEVEL_READ_COMMITTED as READ_COMMITTED
 except ImportError:
@@ -189,7 +188,7 @@ except ImportError:
 
 from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils.six import iteritems
-
+from ansible_collections.community.postgresql.plugins.module_utils.version import LooseVersion
 from ansible_collections.community.postgresql.plugins.module_utils.database import (
     check_input,
 )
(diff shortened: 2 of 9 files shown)


plugins/modules/postgresql_tablespace.py
```diff
@@ -437,8 +437,8 @@ def main():
 
     # Change autocommit to False if check_mode:
     if module.check_mode:
-        if PSYCOPG2_VERSION >= '2.4.2':
-            db_connection.set_session(autocommit=False)
+        if PSYCOPG_VERSION >= LooseVersion("2.4.2"):
+            db_connection.autocommit = False
```
Andersson007, line 441:
```suggestion
        if PSYCOPG_VERSION >= LooseVersion("2.4.2"):
            db_connection.autocommit = False
```
@elpavel as far as i can see `connect_to_db`  now cares about all that stuff. Should we maybe change the line a bit above:
```
db_connection, dummy = connect_to_db(module, conn_params, autocommit=True)
```
to something like:
```
db_connection, dummy = connect_to_db(module, conn_params, autocommit=True if not module.check_mode else False)
```
I'm not sure if this exact syntax works though, but the idea should be clear

Andersson007: ah, i see now that it repeats:) so maybe also a separate function just to reduce code repetition?

changelogs/fragments/533-postgresql_tablespace.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- "postgresql_tablespace - improved autocommit handling (https://github.com/ansible-collections/community.postgresql/pull/533)."
```
Andersson007, line 2:
```suggestion
minor_changes:
- "postgresql_tablespace - improved autocommit handling (https://github.com/ansible-collections/community.postgresql/pull/533)."
```
as this is just refactoring, could you please remove the fragment?

plugins/modules/postgresql_tablespace.py
Andersson007: I'm not insisting. If you think that now is better, please just mark this as solved

Andersson007 (approved):

plugins/modules/postgresql_tablespace.py
elpavel: Separate function would make sense, as we use the same logic three times. The function could be moved to postgres.py, to keep the psycopg logic there.

Andersson007: would you like to do it here or later (which would be also fine) ?

elpavel: I will do it here.

elpavel:
Sanity checks failing for 2.16 with
```
Running sanity test "ignores"
ERROR: Found 1 ignores issue(s) which need to be resolved:
ERROR: tests/sanity/ignore-2.16.txt:1:1: Sanity test 'replace-urlopen' does not test path 'tests/utils/shippable/check_matrix.py' (0%)
See documentation for help: https://docs.ansible.com/ansible-core/devel/dev_guide/testing/sanity/ignores.html
```
any ideas what this means?

Andersson007:
> Sanity checks failing for 2.16 with
> 
> ```
> Running sanity test "ignores"
> ERROR: Found 1 ignores issue(s) which need to be resolved:
> ERROR: tests/sanity/ignore-2.16.txt:1:1: Sanity test 'replace-urlopen' does not test path 'tests/utils/shippable/check_matrix.py' (0%)
> See documentation for help: https://docs.ansible.com/ansible-core/devel/dev_guide/testing/sanity/ignores.html
> ```
> 
> any ideas what this means?

@elpavel i commented in that PR about urlopen, thanks!

jchancojr (approved):

hunleyd (approved):

PR #536, opened 2023-07-14
postgresql_ping test: use vars instead of defaults for test variables

Andersson007:
##### SUMMARY
postgresql_ping test: use vars instead of defaults for test variables as we don't expect them to be overwritten
https://docs.ansible.com/ansible/2.9/user_guide/playbooks_variables.html#examples-of-where-to-set-a-variable


hunleyd (approved):

Andersson007:
@hunleyd thanks for reviewing, merging and backporting!

PR #548, opened 2023-08-02
Cursor query refactor

elpavel:
##### SUMMARY
The `cursor.query` attribute is no longer available in Psycopg 3. It could be replaced with `cursor._query`, but that comes with a [large warning](https://www.psycopg.org/psycopg3/docs/api/cursors.html#psycopg.Cursor._query).

Using `cursor.mogrify()` instead to return the last executed query, it works well in both Psycopg 2 and 3.
This is a small patch, but an important one, helping to unblock #499.

##### ISSUE TYPE
- Refactoring for Psycopg 3

##### COMPONENT NAME
 - postgresql_query
 - postgresql_script

##### ADDITIONAL INFORMATION
Not sure about the variable naming, let me know if you have any suggestions.

### DIFF
--- a/plugins/modules/postgresql_query.py
+++ b/plugins/modules/postgresql_query.py
@@ -261,6 +261,8 @@ except ImportError:
 import re
 
 from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils._text import to_native
+from ansible.module_utils.six import iteritems
 from ansible_collections.community.postgresql.plugins.module_utils.database import (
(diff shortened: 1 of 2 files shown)


hunleyd:
does this need a changelog fragment @Andersson007 ?

Andersson007:
> does this need a changelog fragment @Andersson007 ? i assume not

@hunleyd yeah, let's not backport:)

PR #549, opened 2023-08-03
Additional Psycopg tests

elpavel:
##### SUMMARY
Modified existing tests to work with Psycopg 3 and added extra Psycopg related tests.

The postgresql_user module handles the "database is read only" error, although only when altering a user. Tests were added to verify the behaviour.

Tests were added to verify the path in postgresql_query when `encoding` is specified.

`query_result` should be a list. When the list is empty, currently using psycopg2 we return an empty dict. When psycopg3 is used the same code will return an empty list. I think it is OK to keep the current behaviour, but no need to preserve that for psycopg3.

##### ISSUE TYPE
- Additional tests

##### COMPONENT NAME
Test harness

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/tests/integration/targets/postgresql_query/tasks/postgresql_query_initial.yml
+++ b/tests/integration/targets/postgresql_query/tasks/postgresql_query_initial.yml
@@ -41,8 +41,8 @@
       - result.query_list == ['ANALYZE test_table']
       - result.rowcount == 0
       - result.statusmessage == 'ANALYZE'
-      - result.query_result == {} or result.query_result == []  # Psycopg2 or Psycopg 3
-      - result.query_all_results == [{}] or result.query_all_results == [[]]  # Psycopg2 or Psycopg 3
+      - result.query_result == {}
+      - result.query_all_results == [{}]
 
   - name: postgresql_query - simple select query to test_table
     become_user: '{{ pg_user }}'
@@ -121,7 +121,7 @@
       - result.query == "UPDATE test_table SET story = 'new' WHERE id = 3"
       - result.rowcount == 1
       - result.statusmessage == 'UPDATE 1'
-      - result.query_result == {} or result.query_result == []  # Psycopg2 or Psycopg 3
+      - result.query_result == {}
 
   - name: check the previous update
     become_user: '{{ pg_user }}'
@@ -150,7 +150,7 @@
       - result.query == "UPDATE test_table SET story = 'CHECK_MODE' WHERE id = 3"
       - result.rowcount == 1
       - result.statusmessage == 'UPDATE 1'
-      - result.query_result == {} or result.query_result == []  # Psycopg2 or Psycopg 3
+      - result.query_result == {}
 
   - name: check the previous update that nothing has been changed
     become_user: '{{ pg_user }}'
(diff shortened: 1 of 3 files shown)


tests/integration/targets/postgresql_query/tasks/postgresql_query_initial.yml
```diff
@@ -41,8 +41,8 @@
       - result.query_list == ['ANALYZE test_table']
       - result.rowcount == 0
       - result.statusmessage == 'ANALYZE'
-      - result.query_result == {}
-      - result.query_all_results == [{}]
+      - result.query_result == {} or result.query_result == []  # Psycopg2 or Psycopg 3
+      - result.query_all_results == [{}] or result.query_all_results == [[]]  # Psycopg2 or Psycopg 3
```
Andersson007, line 45: Do you think if it's impossible to convert returned rows to dictionaries in Psycopg 3?
```
query_result:
    description:
    - List of dictionaries in column:value form representing returned rows.
    - When running queries from a file, returns result of the last query.
    returned: success
    type: list
    elements: dict
    sample: [{"Column": "Value1"},{"Column": "Value2"}]
```

Andersson007: asking as it'd be a breaking change as users can rely on checking if the result is empty or not

Andersson007: so i would suggest converting

elpavel: I think both versions return list of dictionaries, when there is data to return.
The only difference in behaviour is when there is no result set.

What should be returned in query_result when there is nothing to return? Empty list or empty dictionary? 
Using Psycopg 3 we get an empty list `[]` which may be the correct option.

The options would be to 
a) fix the existing behaviour 
b) keep the existing behaviour 
c) accept existing behaviour but fix it for psycopg3.

Andersson007: i think it should continue to return what it returns now when empty, i.e. `[{}]`, until there are reasonable concerns raised
my opinion:)

elpavel: ok, lets keep the existing behaviour

Andersson007 (approved):
LGTM

tests/integration/targets/postgresql_query/tasks/postgresql_query_initial.yml
```diff
@@ -234,6 +234,37 @@
...
+    postgresql_query:
+      <<: *pg_parameters
+      query: ALTER TABLE test_table ADD COLUMN foo888 int
+      encoding: 'UTF-888-bad'
+    register: result
+    ignore_errors: true
+
+  - assert:
+      that:
+      - result.failed == true
```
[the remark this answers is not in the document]
elpavel: No worries, I just copy/pasted to keep the tests consistent with the rest of the file.

hunleyd (approved):

hunleyd:
do additional tests go to back branches @Andersson007 ?

Andersson007:
> do additional tests go to back branches @Andersson007 ?

i guess if it's about Psycopg 3 it's fine not to backport

PR #600, opened 2023-10-09
Clean CI because Ansible will drop python2.7 & 3.6

l00ptr:
##### SUMMARY

Pretty soon Ansible will drop support for python2.7 and 3.6, so let's remove some old system from our CI (CentOS7 and RHEL8.8).

We also now try on RHEL8.9

### DIFF
--- a/.azure-pipelines/azure-pipelines.yml
+++ b/.azure-pipelines/azure-pipelines.yml
@@ -143,8 +143,6 @@ stages:
         parameters:
           testFormat: devel/linux/{0}/1
           targets:
-            - name: CentOS 7
-              test: centos7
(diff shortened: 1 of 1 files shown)


l00ptr:
so we should probably clean some old compat code for python 2.7 after merging this one. And we will be able to introduce black :)

hunleyd (changes requested):
why are we dropping older than 2.16 for remote but leaving them for the other tests? and don't the 'summary' lines need to be updated as well (line 247+)

Andersson007:
closed via https://github.com/ansible-collections/community.postgresql/pull/605

PR #611, opened 2023-11-07
Reformat and test with black

l00ptr:
Since we will drop python2.7 we can start using black


Andersson007:
we shouldn't drop support for Python 2.7 as it's still supported by core on target hosts

hunleyd:
> Drop test for EOL CentOS7

C7/R7 isn't EOL until Aug 2024

PR #615, opened 2023-11-09
Ignore sanity errors

Andersson007:
##### SUMMARY

Ignore sanity failings

Spotted the red shield in README

if i change `type` to `isinstance`, it fails with another error that the second type is not valid, so let's ignore it as it's just "not idiomatic".


Andersson007:
ready for review

hunleyd (approved):

Issue #614, opened 2023-11-09
Comment on db objects?


Andersson007:
### Discussed in https://github.com/ansible-collections/community.postgresql/discussions/352

<div type='discussions-op-text'>

<sup>Originally posted by **andymel123** October  8, 2022</sup>
Hi,

is there no way to add comments on db objects?

When I add a db, schema, table,... I would like to also add some documentary comments on that object.

For example in sql I would do something like this:
```
CREATE SCHEMA xy_management;
COMMENT ON SCHEMA xy_management IS 'We use this schema to ensure objects that xy-customers can change are grouped to ...';
```

Is there a way to add such a comment with `postgresql_schema`?</div>

Andersson007:
![Screenshot from 2023-11-09 08-28-43](https://github.com/ansible-collections/community.postgresql/assets/34477873/0c911b05-4406-4b1d-9b8f-716ec70c2b63)

![Screenshot from 2023-11-09 08-28-55](https://github.com/ansible-collections/community.postgresql/assets/34477873/09b43f13-e7c5-404c-b084-a89bd146e5c6)

Andersson007:
@andymel123 @betanummeric @hunleyd let's continue here. We'll disable the Discussions feature in favor of the Forum soon

Andersson007:
ah, now i see a similar issue was created https://github.com/ansible-collections/community.postgresql/issues/354, but let's keep this one for more context

Andersson007:
closed via https://github.com/ansible-collections/community.postgresql/issues/354

PR #646, opened 2023-12-10
postgresql_db: add comment feature

Andersson007:
##### SUMMARY
Relates to https://github.com/ansible-collections/community.postgresql/issues/614
Adds the comment argument to the postgresql_db module

### DIFF
--- a/tests/integration/targets/postgresql_db/tasks/postgresql_db_comment.yml
+++ b/tests/integration/targets/postgresql_db/tasks/postgresql_db_comment.yml
@@ -162,7 +162,7 @@
 - name: Check the comments match
   assert:
     that:
-      - result.query_result[0]['comment'] == null
+      - result.query_result[0]['comment'] == None
 
 
 - name: Clean up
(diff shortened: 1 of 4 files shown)


Andersson007:
ready for review

tests/integration/targets/postgresql_db/tasks/postgresql_db_comment.yml
```diff
@@ -0,0 +1,173 @@
...
+    state: present
+    name: comment_db
+    login_user: "{{ pg_user }}"
+    comment: Another comment
+  check_mode: true
+
+- name: Assert the result
+  assert:
+    that:
+       - result is changed
```
hunleyd, line 51: wouldn't this be 'result is not changed' in check mode?

hunleyd (changes requested):

tests/integration/targets/postgresql_db/tasks/postgresql_db_comment.yml
Andersson007: how so?:) if a passed comment is different from what we currently have in the db, it should report changed yes even in check mode, otherwise it'd be misleading. The check mode should show same result as in real mode not doing actual changes

Issue #354, opened 2022-10-12
have a `comment` argument wherever an object is created


hunleyd:
From a user perspective I think it would be more convenient to have a `comment` argument wherever an object is created. To only change a comment, they need to be idempotent. At a quick glance, the argument would suit 11 modules:

- postgresql_db
- postgresql_ext
- postgresql_idx
- postgresql_lang
- postgresql_publication
- postgresql_schema
- postgresql_sequence
- postgresql_subscription
- postgresql_table
- postgresql_tablespace
- postgresql_user

I'd say the benefit of a `comment` argument or `postgresql_comment` module over `postgresql_query` is that we can validate the kind of object which is commented, to avoid SQL injections. As this is part of the query syntax, not a variable, we cannot rely on psycopg2/3 variable escaping.

_Originally posted by @betanummeric in https://github.com/ansible-collections/community.postgresql/discussions/352#discussioncomment-3834647_

hunleyd:
> the comment argument feels like a good thing to me, it must be idempotent, of course.
it can be a shared class/function

Andersson007:
for postgresql_db https://github.com/ansible-collections/community.postgresql/pull/646

PR #653, opened 2023-12-18
postgresql_db,_schema,_tablespace,_user: add comment value to check_input function

Andersson007:
##### SUMMARY
Relates to https://github.com/ansible-collections/community.postgresql/issues/354
Forget to do it while implementing the comment argument feature.
Note: Not adding this to every module's tests as the function used is the same. In case it changes, the whole set of test targets will be triggered anyway, so having `trust_input: false` only in one target is imo enough.

### DIFF
--- a/plugins/modules/postgresql_db.py
+++ b/plugins/modules/postgresql_db.py
@@ -706,7 +706,8 @@ def main():
     # Check input
     if not trust_input:
         # Check input for potentially dangerous elements:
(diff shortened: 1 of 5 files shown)


Andersson007:
ready for review

hunleyd (approved):
no changelog fragment?

Andersson007:
> no changelog fragment?

nope. it should've been done in those PRs that added the feature. This hasn't been released yet, so no fragment is needed

Andersson007:
thanks for reviewing and merging!

PR #722, opened 2024-07-17
Users configuration

### LINKED ISSUE #598: Set role or database parameters
### Discussed in https://github.com/ansible-collections/community.postgresql/discussions/576

<div type='discussions-op-text'>

<sup>Originally posted by **ivanov17** September  5, 2023</sup>
Hello! What is the best way to set role or database parameters?

There is the `community.postgresql.postgresql_set` module that uses ALTER SYSTEM to set PostgreSQL server configuration parameters.

Unfortunately, I don't see anything like this for altering role or database parameters.

Is using SQL queries with `community.postgresql.postgresql_query` currently the only option for this? Or maybe there are some other options that I didn't notice?

Are there plans to implement other `*_set` modules such as `postgresql_user_set` and `postgresql_db_set`?</div>

toydarian:
##### SUMMARY
PostgreSQL has the ability to set defaults for configuration parameters for roles. This means, that when that users logs in, the parameter for their session will be set to the default for their role instead of the global default.  
For example `work_mem` could be set to a different default for a specific role.  
The current state of the collection doesn't handle this in a good way, though it is possible to manage those using the `community.postgresql.postgresql_query` module. It would be better if those would be handled by the `community.postgresql.postgresql_user` module.  
This is what these changes do. Configuration defaults can be handled when the user is created and will be managed in an idempotent manner. To provide backwards-compatibility, omitting the new parameter on the module will not change the current state.

Personal motivation:  
I need this to manage settings for `pg_audit` for different users. But I think having this in the module is valuable enough, so I added it.  
I also found a ticket on this.

Fixes #598 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`community.postgresql.postgresql_user`

##### ADDITIONAL INFORMATION
I don't have more information than what is already mentioned in the summary.

### DIFF
--- a/changelogs/fragments/722-user-configuration.yml
+++ b/changelogs/fragments/722-user-configuration.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - postgres - adds a feature to 'postgresql_user' that allows to manage user-specific default configuration
--- a/changelogs/fragments/722-user-configuration.yml
+++ b/changelogs/fragments/722-user-configuration.yml
@@ -1,3 +1,4 @@
 ---
 minor_changes:
-  - postgres - adds a feature to 'postgresql_user' that allows to manage user-specific default configuration
+  - postgresql_user - adds the ``configuration`` argument that allows to manage user-specific default configuration
+    (https://github.com/ansible-collections/community.postgresql/issues/598).
--- a/plugins/modules/postgresql_user.py
+++ b/plugins/modules/postgresql_user.py
@@ -951,6 +951,7 @@ def add_comment(cursor, user, comment, check_mode):
 
 
 def compare_user_configurations(current, desired):
+    """Compares two configurations and returns a list of values to reset as well as a dict of parameters to update."""
     reset = []
     update = desired.copy()
 
@@ -969,10 +970,16 @@ def compare_user_configurations(current, desired):
     return {"reset": reset, "update": update}
 
 
-def parse_user_configuration(configs):
+def parse_user_configuration(module, configs):
+    """Parses configuration from a list of 'key=value' strings like returned from the database to a dict."""
     if configs is not None:
-        # parses a list of "key=value" strings to a dict
-        return {t[0]: t[1] for t in map(lambda s: s.split("=", 1), configs)}
+        try:
+            # parses a list of "key=value" strings to a dict
+            return {t[0]: t[1] for t in map(lambda s: s.split("=", 1), configs)}
+        except IndexError:
+            module.fail_json(
+                msg="The 'configuration' option needs to contain a list of strings where each string "
+                "has the format 'key=value'.")
     else:
         return {}
 
@@ -987,7 +994,7 @@ def user_configuration(cursor, module, user, configuration):
     if current_config is None:
         module.fail_json(msg="Can't find user %(user)s in 'pg_roles'" % {"user": user})
 
-    current_config_dict = parse_user_configuration(current_config['rolconfig'])
+    current_config_dict = parse_user_configuration(module, current_config['rolconfig'])
     config_updates = compare_user_configurations(current_config_dict, configuration)
 
     try:
@@ -1116,13 +1123,8 @@ def main():
             if configuration[0].lower() == "purge":
                 changed = user_configuration(cursor, module, user, {}) or changed
             else:
-                try:
-                    config = parse_user_configuration(configuration)
-                    changed = user_configuration(cursor, module, user, config) or changed
-                except IndexError:
-                    module.fail_json(
-                        msg="The 'configuration' option needs to contain a list of strings where each string "
-                            "has the format 'key=value'.")
+                config = parse_user_configuration(module, configuration)
+                changed = user_configuration(cursor, module, user, config) or changed
 
     else:
         if user_exists(cursor, user):
--- a/plugins/modules/postgresql_user.py
+++ b/plugins/modules/postgresql_user.py
@@ -304,13 +304,14 @@ EXAMPLES = r'''
     configuration:
       - work_mem=16MB
 
-# Make sure all configuration is purged for a user
+# Make sure user has only specified default configuration parameters
 - name: Clear all configuration for user
   community.postgresql.postgresql_user:
     name: appclient
     password: "secret123"
     configuration:
-      - purge
+      - work_mem=16MB
+    purge_unspecified_configuration: true
 '''
 
 RETURN = r'''
--- a/plugins/modules/postgresql_user.py
+++ b/plugins/modules/postgresql_user.py
@@ -1105,6 +1105,12 @@ def main():
 
     srv_version = get_server_version(db_connection)
 
+    # sanitize configuration
+    for key, value in configuration.items():
+        if '"' in key or '\'' in key:
+            module.fail_json("The key of a configuration may not contain single or double quotes")
+        configuration[key] = value.replace("'", "''")
+
     try:
         role_attr_flags = parse_role_attrs(role_attr_flags, srv_version)
     except InvalidFlagsError as e:
--- a/plugins/modules/postgresql_user.py
+++ b/plugins/modules/postgresql_user.py
@@ -162,12 +162,12 @@ options:
       - Removes any parameter from the user that is not listed here.
       - If the first item is 'purge', all configuration parameters will be removed from the user.
       - If omitted or empty, the current parameters of the user will not be changed.
-      - Using this will leave the I(user) as well as this parameter open to SQL-injections,
+      - Using this will leave the O(user) as well as this parameter open to SQL-injections,
         parameters are checked if I(trust_input) is C(false).
     type: list
     elements: str
     default: []
-    version_added: '4.0.0'
+    version_added: '3.5.0'
 notes:
 - The module creates a user (role) with login privilege by default.
   Use C(NOLOGIN) I(role_attr_flags) to change this behaviour.
@@ -950,49 +950,58 @@ def add_comment(cursor, user, comment, check_mode):
         return False
 
 
+def compare_user_configurations(current, desired):
+    reset = []
+    update = desired.copy()
+
+    # check each item in the current configuration
+    for key, value in current.items():
+        # we already have the correct setting
+        if key in desired and value == desired[key]:
+            # so we can remove it from the list
+            del update[key]
+        # if the key is not in the list of settings we want
+        elif key not in desired:
+            # we will reset it on the database
+            reset.append(key)
+        # if the setting is not in the db or has the wrong value, it will get updated
+
+    return {"reset": reset, "update": update}
+
+
+def parse_user_configuration(configs):
+    if configs is not None:
+        # parses a list of "key=value" strings to a dict
+        return {t[0]: t[1] for t in map(lambda s: s.split("=", 1), configs)}
+    else:
+        return {}
+
+
 def user_configuration(cursor, module, user, configuration):
     """Updates the user's configuration parameters if necessary."""
     current_config_query = "SELECT rolconfig FROM pg_roles WHERE rolname = %(user)s;"
     cursor.execute(current_config_query, {"user": user})
     current_config = cursor.fetchone()
-    current_config_dict = {}
     changed = False
 
     if current_config is None:
         module.fail_json(msg="Can't find user %(user)s in 'pg_roles'" % {"user": user})
 
-    if current_config['rolconfig'] is not None:
-        for item in current_config['rolconfig']:
-            tokens = item.split("=", 1)
-            current_config_dict[tokens[0]] = tokens[1]
-
-    c = configuration.copy()
-    reset = []
-
-    # check each item in the current configuration
-    for key, value in current_config_dict.items():
-        # we already have the correct setting
-        if key in c and value == c[key]:
-            # so we can remove it from the list
-            del c[key]
-        # if the key is not in the list of settings we want
-        elif key not in c:
-            # we will reset it on the database
-            reset.append(key)
-        # if the setting is not in the db or has the wrong value, it will get updated
+    current_config_dict = parse_user_configuration(current_config['rolconfig'])
+    config_updates = compare_user_configurations(current_config_dict, configuration)
 
     try:
-        # It seems psycopg's prepared statements don't work with 'ALTER ROLE' as this point.
+        # It seems psycopg's prepared statements don't work with 'ALTER ROLE' at this point.
         # This is vulnerable to SQL-injections (added to docs) but I don't see a better way to do this.
-        for item in reset:
+        for item in config_updates["reset"]:
             query = 'ALTER ROLE "%(user)s" RESET "%(key)s";' % {"user": user, "key": item}
-            cursor.execute(query)
             executed_queries.append(query)
+            cursor.execute(query)
             changed = True
-        for key, value in c.items():
+        for key, value in config_updates["update"].items():
             query = 'ALTER ROLE "%(user)s" SET "%(key)s" TO \'%(value)s\';' % {"user": user, "key": key, "value": value}
-            cursor.execute(query)
             executed_queries.append(query)
+            cursor.execute(query)
             changed = True
     except psycopg.ProgrammingError as e:
         module.fail_json("Unable to update configuration for '%(user)s' due to: %(exception)s" %
@@ -1021,7 +1030,7 @@ def main():
         session_role=dict(type='str'),
         comment=dict(type='str', default=None),
         trust_input=dict(type='bool', default=True),
-        configuration=dict(type='list', elements='str', default=[])
+        configuration=dict(type='list', elements='str', default=[]),
     )
     module = AnsibleModule(
         argument_spec=argument_spec,
@@ -1108,8 +1117,7 @@ def main():
                 changed = user_configuration(cursor, module, user, {}) or changed
             else:
                 try:
-                    # parses a list of "key=value" strings to a dict
-                    config = {t[0]: t[1] for t in map(lambda s: s.split("=", 1), configuration)}
+                    config = parse_user_configuration(configuration)
                     changed = user_configuration(cursor, module, user, config) or changed
                 except IndexError:
                     module.fail_json(
--- a/plugins/modules/postgresql_user.py
+++ b/plugins/modules/postgresql_user.py
@@ -157,16 +157,23 @@ options:
   configuration:
     description:
       - Role-specific configuration parameters that would otherwise be set by C(ALTER ROLE user SET variable TO value;).
-      - Takes a list of strings like C(key=value) that is split on the first equal-sign.
+      - Takes a dict where the key is the name of the configuration parameter. If the key includes special characters
+        like C(.) and C(-), it needs to be quoted to ensure the YAML is valid.
       - Sets or updates any parameter in the list that is not present or has the wrong value in the database.
       - Removes any parameter from the user that is not listed here.
-      - If the first item is 'purge', all configuration parameters will be removed from the user.
-      - If omitted or empty, the current parameters of the user will not be changed.
+      - If omitted or empty, the current configuration of that user will only be purged if
+        O(purge_on_empty_configuration=true).
       - Using this will leave the O(user) as well as this parameter open to SQL-injections,
         parameters are checked if I(trust_input) is C(false).
-    type: list
-    elements: str
-    default: []
+    type: dict
+    default: {}
+    version_added: '3.5.0'
+  purge_on_empty_configuration:
+    description:
+      - If set to C(true), the user's default configuration parameters will be purged in case O(configuration) is empty
+        or omitted, otherwise existing parameters will not be modified if O(configuration) is empty.
+    type: bool
+    default: false
     version_added: '3.5.0'
 notes:
 - The module creates a user (role) with login privilege by default.
@@ -978,8 +985,7 @@ def parse_user_configuration(module, configs):
             return {t[0]: t[1] for t in map(lambda s: s.split("=", 1), configs)}
         except IndexError:
             module.fail_json(
-                msg="The 'configuration' option needs to contain a list of strings where each string "
-                "has the format 'key=value'.")
+                msg="Expecting a list of strings where each string has the format 'key=value'.")
     else:
         return {}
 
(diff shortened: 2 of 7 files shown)


plugins/modules/postgresql_user.py
```diff
@@ -154,6 +154,20 @@
...
+      - Sets or updates any parameter in the list that is not present or has the wrong value in the database.
+      - Removes any parameter from the user that is not listed here.
+      - If the first item is 'purge', all configuration parameters will be removed from the user.
+      - If omitted or empty, the current parameters of the user will not be changed.
+      - Using this will leave the I(user) as well as this parameter open to SQL-injections,
+        parameters are checked if I(trust_input) is C(false).
+    type: list
+    elements: str
+    default: []
+    version_added: '4.0.0'
```
Andersson007, line 170:
```suggestion
    version_added: '3.5.0'
```
it's a minor change as it doesn't break the interface

plugins/modules/postgresql_user.py
```diff
@@ -274,6 +288,22 @@
...
     name: monitoring
     priv: 'pg_catalog.pg_stat_database:SELECT'
+
+# Create a user and set a default-configuration that is active when they start a session
+- name: Create a user with config-parameter
+  community.postgresql.postgresql_user:
+    name: appclient
+    password: "secret123"
+    configuration:
+      - work_mem=16MB
```
Andersson007, line 298:
```suggestion
    configuration:
      - work_mem=16MB
```
How about making it a dict instead, it'd look like:
```
    configuration:
      work_mem: 16MB
      purge: True
```

plugins/modules/postgresql_user.py
```diff
@@ -920,6 +950,56 @@ def add_comment(cursor, user, comment, check_mode):
...
+    current_config_dict = {}
+    changed = False
+
+    if current_config is None:
+        module.fail_json(msg="Can't find user %(user)s in 'pg_roles'" % {"user": user})
+
+    if current_config['rolconfig'] is not None:
+        for item in current_config['rolconfig']:
+            tokens = item.split("=", 1)
+            current_config_dict[tokens[0]] = tokens[1]
```
Andersson007, line 967:
```suggestion
    if current_config['rolconfig'] is not None:
        for item in current_config['rolconfig']:
            tokens = item.split("=", 1)
            current_config_dict[tokens[0]] = tokens[1]
```
i would suggest making it a dict as mentioned above and move this to a separate function. You could cover it with unit tests then if you wish, though only integration tests are mandatory.

plugins/modules/postgresql_user.py
```diff
@@ -920,6 +950,56 @@ def add_comment(cursor, user, comment, check_mode):
...
+    # check each item in the current configuration
+    for key, value in current_config_dict.items():
+        # we already have the correct setting
+        if key in c and value == c[key]:
+            # so we can remove it from the list
+            del c[key]
+        # if the key is not in the list of settings we want
+        elif key not in c:
+            # we will reset it on the database
+            reset.append(key)
```
Andersson007, line 981:
```suggestion
    # check each item in the current configuration
    for key, value in current_config_dict.items():
        # we already have the correct setting
        if key in c and value == c[key]:
            # so we can remove it from the list
            del c[key]
        # if the key is not in the list of settings we want
        elif key not in c:
            # we will reset it on the database
            reset.append(key)
```
would move this to a separate function too and cover it with unit tests

plugins/modules/postgresql_user.py
```diff
@@ -920,6 +950,56 @@ def add_comment(cursor, user, comment, check_mode):
...
+        # if the setting is not in the db or has the wrong value, it will get updated
+
+    try:
+        # It seems psycopg's prepared statements don't work with 'ALTER ROLE' as this point.
+        # This is vulnerable to SQL-injections (added to docs) but I don't see a better way to do this.
+        for item in reset:
+            query = 'ALTER ROLE "%(user)s" RESET "%(key)s";' % {"user": user, "key": item}
+            cursor.execute(query)
+            executed_queries.append(query)
```
Andersson007, line 990:
```suggestion
            executed_queries.append(query)
            cursor.execute(query)
```
i suggest appending first as it can be helpful when debugging, if you don't mind

plugins/modules/postgresql_user.py
```diff
@@ -920,6 +950,56 @@ def add_comment(cursor, user, comment, check_mode):
...
+            changed = True
+        for key, value in c.items():
+            query = 'ALTER ROLE "%(user)s" SET "%(key)s" TO \'%(value)s\';' % {"user": user, "key": key, "value": value}
+            cursor.execute(query)
+            executed_queries.append(query)
```
Andersson007, line 995:
```suggestion
            executed_queries.append(query)
            cursor.execute(query)
```

plugins/modules/postgresql_user.py
```diff
@@ -941,6 +1021,7 @@ def main():
         session_role=dict(type='str'),
         comment=dict(type='str', default=None),
         trust_input=dict(type='bool', default=True),
+        configuration=dict(type='list', elements='str', default=[])
```
Andersson007, line 1024:
```suggestion
        configuration=dict(type='list', elements='str', default=[]),
```

Andersson007 (changes requested):
@till hello, thanks for the very well-made PR! 

Could you also please:
- Add a [changlog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment) so that users know about the chage
- Add equivalents of the integration test tasks with `check_mode: true` and check that the actual state hasn't been changed
- (fully optional) Split that function into several and cover it with unit tests

hunleyd (changes requested):
<3 the idea of this added functionality. pls re-request reviews once you've addressed andersson007's concerns.

plugins/modules/postgresql_user.py
toydarian: I have been thinking about various ways to do this. The dict method has the issue that parameters can include dots (e.g. `pg_audit.log`) which would need to be quoted in YAML which might inconvenience users who forget or don't know that.

I also thought about this, which I find too bulky:
```
configuration:
  - key: work_mem
    value: 16MB
  - key: ...
    value: ...
```
Also I'm not too happy with the way I'm doing it it right now.  
Let me know which one you prefer.

Andersson007: I'm personally in favor of that one i posted even if it needs quoting (we can make a short note about this):
```
    configuration:
      work_mem: 16MB
      purge: True
```
I'm not necessarily against of having it as in the PR now though. The dict form looks better to me but it's subjective. Also we usually handle similar cases with dicts here and there.
@hunleyd @betanummeric what's your feeling?

changelogs/fragments/722-user-configuration.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - postgres - adds a feature to 'postgresql_user' that allows to manage user-specific default configuration
```
Andersson007, line 3:
```suggestion
  - postgresql_user - adds the ``configuration`` argument that allows to manage user-specific default configuration (https://github.com/ansible-collections/community.postgresql/issues/598).
```
thanks! to make the line consistent with other entries

plugins/modules/postgresql_user.py
toydarian: Okay, I changed it in that way, but I added a separate parameter to enable purging on an empty dict, otherwise this can get really confusing

betanummeric: The
```
configuration:
  - work_mem=16MB
```
style means that we need to implement custom parsing and document that to distinguish key and value. What if the key contains a `=` or some other special chars?
I would prefer the
```
configuration:
  work_mem: 16MB
```
style because it uses standard yaml parsing and also prevents duplicate keys at paring level.

Yes, yaml may have some quirks new users may stumble upon, but if we make something custom it's more to learn for everybody.

betanummeric:
I think in some cases it may be useful to only touch variables explicitly specified in the module arguments.
With module arguments for example like
```yaml
set_variables:
  search_path: foo
reset_variables:
- work_mem
reset_unspecified: false
```
would explicitly set `search_path`, explicitly reset `work_mem` and leave all other potentially configured things untouched.

plugins/modules/postgresql_user.py
```diff
@@ -920,6 +957,71 @@ def add_comment(cursor, user, comment, check_mode):
...
+
+    if current_config is None:
+        module.fail_json(msg="Can't find user %(user)s in 'pg_roles'" % {"user": user})
+
+    current_config_dict = parse_user_configuration(module, current_config['rolconfig'])
+    config_updates = compare_user_configurations(current_config_dict, configuration)
+
+    try:
+        # It seems psycopg's prepared statements don't work with 'ALTER ROLE' at this point.
+        # This is vulnerable to SQL-injections (added to docs) but I don't see a better way to do this.
```
betanummeric, line 1008: psycopg is serializing values as _constants_ (numbers, `'`-quoted strings, ...), but `alter role` expects _identifiers_ (`"`-quoted).
This project already dealt with SQL-injection prevention for identifiers, please use the existing [`pg_quote_identifier`](https://github.com/ansible-collections/community.postgresql/blob/ea28b3de22464d419d299722dd250033798886c6/plugins/module_utils/database.py#L125) function for identifiers (role name, config variable name).

psycopg should serialize the config variable value.

Thereby, this can be made safe against SQL injections.

betanummeric:
@toydarian 

> I'm not a big fan of that. If you had `reset_unspecified: false` you would need to list all variables to reset which can be a long list. If at all, I would specify a list of parameters not to reset, but I think, even that is overkill. I think the current implementation strikes a good balance between clarity and covering most use-cases as well as providing backwards-compatibility.

With `reset_unspecified` I meant to support these two modes, which is a common pattern across the collection:
A) set the requested things and reset everything else, ensuring only requested things deviate from defaults
B) set the requested things and leave the rest untouched

Mode A is useful when you have authoritative and comprehensive information for what shall be configured.
Mode B is better when you only know how some parts shall be configured, leaving the rest up to others.

I'm using the collection to the extend possible only in mode B, integrated in some high-level automation. I'm doing this to step-wise introduce more automation to an already-existing database platform.
When starting to use user-config, I would take a handful variables and either set or reset them, without needing to specify "long" lists.

As a user, I'd like to have the choice between these modes.
We can make mode A the default (`reset_unspecified: true`). 

I don't think we need a "don't reset these variables" list argument, this sounds too special.

---

> There are also other use-cases, for example `IN DATABASE` where you can set a default for a role only in a certain named database and then you have `ALTER DATABASE` to set defaults for a specific database, which are also not covered. If I tried to cover all of those, that would probably warrant a completely new module.

`alter role ... in database ...` would be a feature for this `postgresql_user` module. It could be implemented with another argument holding a nested dict:
```yaml
db_config:
  some_specific_database:
    maintenance_work_mem: '16MB'
```
`alter database ...` is a topic for the `postgresql_db` module.

No need to implement either in this PR.

plugins/modules/postgresql_user.py
toydarian: I tried, but [`pg_quote_identifier`](https://github.com/ansible-collections/community.postgresql/blob/ea28b3de22464d419d299722dd250033798886c6/plugins/module_utils/database.py#L125) only allows one dot, while role names, as well as configuration parameters may include more than one dot.  
That is probably also why it is not used in [`user_add`](https://github.com/ansible-collections/community.postgresql/blob/ea28b3de22464d419d299722dd250033798886c6/plugins/modules/postgresql_user.py#L386), for example.

toydarian:
@betanummeric okay, now I understand your use-case. I incorporated the necessary changes.

plugins/modules/postgresql_user.py
betanummeric: Well ok, `pg_quote_identifier` could be better documented. It actually supports both unquoted and already-quoted inputs as well as chained(?) identifiers like `database.schema.table.column`. It has a parameter for how many identifier parts are allowed.

In this case we could also only support unquoted input and escape it like this:
```python3
sql_safe = '"%s"' % (unsafe_input.replace('"', '""'), )
```
As https://www.postgresql.org/docs/current/sql-syntax-lexical.html explains, this should be enough to safely write an identifier.

toydarian: There are several issues:  
- One is in [this `dict`](https://github.com/ansible-collections/community.postgresql/blob/ea28b3de22464d419d299722dd250033798886c6/plugins/module_utils/database.py#L43). It specifies that a `role` may only have one dot and there is no entry for `parameter` or something like that.
- Another one is, that PostgreSQL doesn't recognize a user if quoted in the way `pg_quote_identifier` returns:
  ```
  postgres=# GRANT CONNECT ON DATABASE "postgres" TO "test"."user" ;
  ERROR:  syntax error at or near "."
  LINE 1: GRANT CONNECT ON DATABASE "postgres" TO "test"."user" ;
                                                        ^
  postgres=# GRANT CONNECT ON DATABASE "postgres" TO "test.user" ;
  GRANT
  ```

Also this would still be vulnerable to SQL-injections in the value even if we quoted the user and parameter name.

I'm using `check_input` when `trust_input` is `false`. It couldn't properly check dicts, so I changed it to be able to check lists, tuples and dicts recursively.  
Imho it is good enough to run the input through `check_input` when `trust_input=false`.

toydarian:
I think I addressed all comments now.
@Andersson007 @betanummeric @hunleyd

plugins/modules/postgresql_user.py
hunleyd: > Okay, I changed it in that way, but I added a separate parameter to enable purging on an empty dict, otherwise this can get really confusing

did you push this change @toydarian ? i don't see a difference here

hunleyd (changes requested):

toydarian:
@hunleyd seems I forgot to update the example, fixed in [3f397d6](https://github.com/ansible-collections/community.postgresql/pull/722/commits/3f397d6732549726fc1459ce2a91df1c710ec99a)

plugins/modules/postgresql_user.py
betanummeric: `_PG_IDENTIFIER_TO_DOT_LEVEL` specifies how many dot-separated parts are allowed, which is one more than how many dots are allowed. For role names, no dots are allowed. Role names are single identifiers, nothing schema-qualified or whatnot.

I think the `check_input` function doesn't help much because it only checks for some specific SQL injections. This is no guarantee to be injection-safe and gives a false sense of trust.

The only way to get protected against SQL injections is by proper input escaping. This is very simple: just double every occurrence of `"`, just as I suggested in the comment earlier.

`pg_quote_identifier` adds some magic around that simple escaping to detect already-quoted input.
If you disklike that magic, ok, don't use it.
But please implement proper input escaping.

toydarian: > For role names, no dots are allowed.

That is not correct. Dots are allowed, but they don't have the same meaning as, for example, in a table-name, where you have `database.schema.table`. So a username `some.user.name` is a valid username, but you can't quote it like this `"some"."user"."name"`. This is what `pg_quote_identifier` does and it simply doesn't work. This has nothing to do with me wanting or not wanting to use that function.

> The only way to get protected against SQL injections is by proper input escaping. This is very simple: just double every occurrence of ", just as I suggested in the comment earlier.

Sorry, but I consider this a very dangerous belief. Relying on escaping is exactly what made, for example, unicode homoglyph sql-injections possible (I assume it is not possible here, but it is an example where trusting in escaping has been tripping people before). Escaping is imho not fit to reliably prevent SQL injections and especially writing custom escaping-functions is prone to oversight and is what actually gives you the false sense of security.  
The "better" (and imho only) way to prevent SQL injections is prepared statements (which doesn't work in this case).
Here is, by the way, what happens when you would do the double-`"` thing in psql (I find this funny):
```
postgres=# create user some""user; select 1;--";
ERROR:  syntax error at or near "some"
LINE 1: create user some""user;
                    ^
 ?column?
----------
        1
(1 row)
```
The first statement throws an error and the second one still gets executed. I wasn't able to reproduce this with psycopg, though.


In any case, you are the reviewer, so if you want it done the way you write, I have no problem doing it that way.  
There is, though, the issue, that this quoting isn't done anywhere else in this module (for example [here](https://github.com/ansible-collections/community.postgresql/blob/ea28b3de22464d419d299722dd250033798886c6/plugins/modules/postgresql_user.py#L386)).  
This leads to the following problems:
If I would only add that to my code, other points (like the one in `user_add`) would still be vulnerable.
I could add this quoting in all places throughout the module (Idk if that is in the scope of this MR), but in that case, we would break backwards-compatibility.  
Somebody who already uses the double-`"` method to create users with quotes might be using `user: 'some""user'` to create a user `some"user`. But if I now start doing the replace-thing, this would lead to the user being created turn into `some""user`, as the input `'some""user'` would get turned into `some""""user`.  
If I only add the replacing to my code, the module will break, as in the existing code, the user would be `some"user` but in my code, it would be `some""user`.  
I could use a regex to only replace single instances of `"` but in that case, we will be vulnerable to SQL injections again, if somebody does `user: 'someuser"""; drop database postgres;--'`.

Bottom line:  
I don't see a consistent way to do this without breaking backwards-compatibility.

My opinion:  
This kind of escaping hasn't been done until now, and if the plan is to do it, it should be done properly, as in a full refactoring where it is done throughout the whole module, but that would probably need to be done in a major-release as it will backwards-compatibility as shown above.

betanummeric: > For role names, no dots are allowed.

Yes, PostgreSQL allows dots in role names. What I wanted to say: For role names, `pg_quote_identifier` only allows a single identifiers, not multiple dot-separated identifiers.
For example:
```python3
>>> pg_quote_identifier('a', 'role')
'"a"'
>>> pg_quote_identifier('"a"', 'role')
'"a"'
>>> pg_quote_identifier('a.b', 'role')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "database.py", line 128, in pg_quote_identifier
    raise SQLParseError('PostgreSQL does not support %s with more than %i dots' % (id_type, _PG_IDENTIFIER_TO_DOT_LEVEL[id_type]))
quote.SQLParseError: PostgreSQL does not support role with more than 1 dots
>>> pg_quote_identifier('"a.b"', 'role')
'"a.b"'
>>> pg_quote_identifier('a"', 'role')
'"a"""'
>>> pg_quote_identifier('💣', 'role')
'"💣"'
```

betanummeric: The example `create user some""user; select 1;--";` doesn't apply because proper escaping would add surrounding quotes in addition to doubling quotes from the input:
```sql
postgres=# create user "some""user; select 1;--";
CREATE ROLE
postgres=# select rolname from pg_roles where rolname like 'some%';
        rolname         
------------------------
 some"user; select 1;--
```

betanummeric: > This kind of escaping hasn't been done until now, and if the plan is to do it, it should be done properly, as in a full refactoring where it is done throughout the whole module, but that would probably need to be done in a major-release as it will backwards-compatibility as shown above.

Escaping by doubling quotes has been done [here](https://github.com/ansible-collections/community.postgresql/blob/ea28b3de22464d419d299722dd250033798886c6/plugins/modules/postgresql_privs.py#L797) and [here](https://github.com/ansible-collections/community.postgresql/blob/ea28b3de22464d419d299722dd250033798886c6/plugins/module_utils/database.py#L106). The latter is part of `pg_quote_identifier`.

Yes, usage is not so consistent throughout the collection and handling all identifier input equal would be ideal.
Yes, making it consistent would require a major release.
I would like that but I don't think this is urgent. -> separate issue

For this PR I would like to see some input escaping instead of none. Preferably with one of the [two methods I described earlier](https://github.com/ansible-collections/community.postgresql/pull/722#discussion_r1688636681).

toydarian: Okay, now I understand, that the role-name needs to be "pre-quoted" before `pg_quote_identifiers` accepts it. I had to change other parts of the code as well to make sure the results are consistent, but I think I managed to get it to work. (see bc873bb)  
I didn't touch the (deprecated) privileges part in the users-module, though, so that is still vulnerable.

I'm also sanitizing the keys and values which should prevent most sql-injections. (see 540a89b)

Andersson007:
@toydarian thanks a lot for working on this!
The unit tests are failing:
```
01:06 _____ ERROR collecting tests/unit/plugins/modules/test_postgresql_user.py ______
01:06 ImportError while importing test module '/root/ansible_collections/community/postgresql/tests/unit/plugins/modules/test_postgresql_user.py'.
01:06 Hint: make sure your test modules/packages have valid Python names.
01:06 Traceback:
01:06 /tmp/ansible-test-b7_edotq/ansible/utils/collection_loader/_collection_finder.py:625: in load_module
01:06     exec(code_obj, module.__dict__)
01:06 tests/unit/plugins/modules/test_postgresql_user.py:12: in <module>
01:06     from plugins.modules.postgresql_user import parse_user_configuration, compare_user_configurations, _pg_quote_user
01:06 E   ImportError: No module named plugins.modules.postgresql_user
01:06 - generated xml file: /root/ansible_collections/community/postgresql/tests/output/junit/python2.7-modules-units.xml -
01:06 =========================== short test summary info ============================
01:06 ERROR tests/unit/plugins/modules/test_postgresql_user.py
```
We should make CI green before merging. As unit tests are optional, in case you have no time to deal with them right now, you could remove them and add later when you have time. So please make CI green however it works for you

toydarian:
@Andersson007 about the unit-tests:  
It seems they only fail for Python 2 and I can't really figure out why. I managed to fix the import-issue but now there is an assertion error and those tests work fine with Python 3. I disabled the failing test for Python 2.  
The CI is green now.

Andersson007 (approved):
@toydarian thanks! +1 from me to merging the PR
@hunleyd @betanummeric any other comments/suggestions or we can merge?

@toydarian is obviously a committed person that i'm sure will be happy to fix all possible reported inconsistencies if there are any;)

hunleyd (approved):

Andersson007:
@toydarian thanks for the contribution!  Much appreciated!
@hunleyd @betanummeric thanks for reviewing!
It was a great long discussion, but we did it!

@toydarian we have a forum group https://forum.ansible.com/g/PostgreSQLTeam and you're welcome to join. Also there's a Matrix channel mentioned in the description  - welcome too if you're interested. There are plenty opportunities to contribute in this and other collections, so please let us know if there's nothing in the issue tracker that look interesting.

PR #437, opened 2023-03-27
Added checking of available columns when receiving a subscription from the database

aleksvagachev:
##### SUMMARY
When an unprivileged user received a subscription, an error occurred: permission denied for table pg_subscription.
Now a check has been added, by getting the available columns and displaying only them.
Solution of the issue: #429 

##### ISSUE TYPE
Feature Pull Request

PR #727, opened 2024-08-28
Handles subscriptions when subconninfo column is unavailable

### LINKED ISSUE #726: postgresql_subscription does not work when pg_subscription.subconninfo column unavailable
##### SUMMARY
The `subconninfo` column of the `pg_subscription` catalog table contains the sensitive connection information needed to establish a logical replication subscription. According to the [documentation](https://www.postgresql.org/docs/current/catalog-pg-subscription.html):
> Access to the column subconninfo is revoked from normal users, because it could contain plain-text passwords.

In Google Cloud Platform's CloudSQL, this column is not accessible at all, as mentioned in their documentation [here](https://cloud.google.com/sql/docs/postgres/replication/configure-logical-replication#limitations-native-logical)
> Limitations on native PostgreSQL logical replication
> Access to the subconninfo column of the [pg_subscription](https://www.postgresql.org/docs/current/static/catalog-pg-subscription.html) system table is unavailable.

This limitation may exist in other cloud providers as well. 

As a result, the `postgresql_subscription` module is unable to manage subscriptions in CloudSQL, as the query of the `pg_subscription` catalog in function `__get_general_subscr_info` fails.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

`postgresql_subscription`

##### ANSIBLE VERSION

```
ansible [core 2.17.1]
  config file = None
  configured module search path = ['/Users/nickwilson/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/nickwilson/.venv/ansible/lib/python3.12/site-packages/ansible
  ansible collection location = /Users/nickwilson/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/nickwilson/.venv/ansible/bin/ansible
  python version = 3.12.5 (main, Aug  6 2024, 19:08:49) [Clang 15.0.0 (clang-1500.3.9.4)] (/Users/nickwilson/.venv/ansible/bin/python3.12)
  jinja version = 3.1.4
  libyaml = True
```

##### COLLECTION VERSION

```
# /Users/nickwilson/.ansible/collections/ansible_collections
Collection           Version
-------------------- -------
community.postgresql 3.5.0

# /Users/nickwilson/.venv/ansible/lib/python3.12/site-packages/ansible_collections
Collection           Version
-------------------- -------
community.postgresql 3.4.1
```

##### CONFIGURATION

```
CONFIG_FILE() = /Users/nickwilson/git/ansible-databases-v2/ansible.cfg
INVENTORY_ENABLED(/Users/nickwilson/git/ansible-databases-v2/ansible.cfg) = ['gcp_compute', 'host_list', 'script', 'auto', 'yaml', 'ini', 'toml', 'advanced_host_list', 'constructed']
PAGER(env: PAGER) = less
```

##### OS / ENVIRONMENT

MacOS 14.6.1 and AWX 21.8.0
This is reproducible on managed databases where we are connecting via host/port, so the OS details are not particularly relevant.

##### STEPS TO REPRODUCE

Create a subscription where the `login_user` does not have access to the `subconninfo` column of `pg_catalog.pg_subscription`
```
    community.postgresql.postgresql_subscription:
      login_user: "testuser"
      login_host: "127.0.0.1"
      db: "ansible_test"
      name: "test_subscription"
      publications: "test_pub"
      connparams:
        host: "127.0.0.1"
        port: "5432"
        user: "sub_user"
        password: "password"
        dbname: "ansible_test2"
      subsparams:
        slot_name: "test_slot"
        create_slot: true
        copy_data: true

```

##### EXPECTED RESULTS
Subscription is created as expected

##### ACTUAL RESULTS

```
The full traceback is:
  File "/var/folders/j0/frglw1v57yx71b11nrtcp24c0000gn/T/ansible_community.postgresql.postgresql_subscription_payload_vbd5cauy/ansible_community.postgresql.postgresql_subscription_payload.zip/ansible_collections/community/postgresql/plugins/module_utils/postgres.py", line 233, in exec_sql
    obj.cursor.execute(query, query_params)
  File "/Users/nickwilson/.venv/ansible/lib/python3.12/site-packages/psycopg2/extras.py", line 146, in execute
    return super().execute(query, vars)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
fatal: [127.0.0.1]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "ca_cert": null,
            "cascade": false,
            "comment": null,
            "connect_params": {},
            "connparams": {
                "dbname": "ansible_test2",
                "host": "127.0.0.1",
                "password": "password",
                "port": "5432",
                "user": "sub_user"
            },
            "db": "ansible_test",
            "login_host": "127.0.0.1",
            "login_password": "",
            "login_unix_socket": "",
            "login_user": "testuser",
            "name": "test_subscription",
            "owner": null,
            "port": 5432,
            "publications": [
                "test_pub"
            ],
            "session_role": null,
            "ssl_cert": null,
            "ssl_key": null,
            "ssl_mode": "prefer",
            "state": "present",
            "subsparams": {
                "copy_data": true,
                "create_slot": true,
                "slot_name": "test_slot"
            },
            "trust_input": true
        }
    },
    "msg": "Cannot execute SQL 'SELECT obj_description(s.oid, 'pg_subscription') AS comment, d.datname, r.rolname, s.subenabled, s.subconninfo, s.subslotname, s.subsynccommit, s.subpublications FROM pg_catalog.pg_subscription s JOIN pg_catalog.pg_database d ON s.subdbid = d.oid JOIN pg_catalog.pg_roles AS r ON s.subowner = r.oid WHERE s.subname = %(name)s AND d.datname = %(db)s': permission denied for table pg_subscription\n"
}
```

nickgsc:
##### SUMMARY

Fixes #726 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`postgresql_subscription`

##### ADDITIONAL INFORMATION

Using https://github.com/ansible-collections/community.postgresql/pull/437 as a bit of an example of how to address this limitation. I've modified the module to first check to see which columns are available (per the information schema) and then adjusting the query accordingly. If the `subconninfo` column is not available, we deal with that by allowing the subscription to be created, but then logging a warning on subsequent runs if we are looking at a subscription that already exists. Without access to `subconninfo`, we have no way to compare the connection info provided to the module versus what exists in the database, so we simply skip.

I am having a hell of a time devising tests that will prove out that this fix works. I've tested it in an actual CloudSQL environment, where there lack of access to the `subconninfo` - but in the integration tests, I haven't been able to devise a way to replicate this restriction because creating a subscription requires superuser access in Postgres versions prior to 16. Using the superuser then bypasses all ACL checks, which renders the test moot (i.e. the problem cannot be reproduced). It should be possible to reproduce and test this with Postgres 16 by using the lesser built-in `pg_create_subscription` role - but the test harness as of now is on PG 15.

Google must have come up with some way to work around this restriction, since they do not give you superuser access yet you can still create subscriptions there using a lesser built-in role they provide, in all versions of Postgres that support logical replication. It's a bit beyond my understanding of Postgres internals to figure out how to replicate Google's configuration in a test scenario. EDIT: Looking at the Postgres source code where the ACL check lives for creating subscriptions, and Google must be running their own slightly customized version of Postgres to support their pseudo-superuser permissions scheme.

Open to any suggestions on how to approach testing this.

```
[WARNING]: 'connparams' is ignored when existing subscription connparms are unknowable
```

### DIFF
--- a/plugins/modules/postgresql_subscription.py
+++ b/plugins/modules/postgresql_subscription.py
@@ -586,15 +586,22 @@ class PgSubscription():
         Returns:
             Dict with subscription information if successful, False otherwise.
         """
+        columns_sub_table = ("SELECT column_name "
+                             "FROM information_schema.columns "
+                             "WHERE table_schema = 'pg_catalog' "
+                             "AND table_name = 'pg_subscription'"
+                             "AND column_name IN ('subenabled','subconninfo','subslotname','subsynccommit','subpublications')")
+        columns_result = exec_sql(self,columns_sub_table,query_params={'name': self.name, 'db': self.db}, add_to_executed=False)
+        columns = ", ".join(["s.%s" % column['column_name'] for column in columns_result])
         query = ("SELECT obj_description(s.oid, 'pg_subscription') AS comment, "
-                 "d.datname, r.rolname, s.subenabled, "
-                 "s.subconninfo, s.subslotname, s.subsynccommit, "
-                 "s.subpublications FROM pg_catalog.pg_subscription s "
+                 "d.datname, r.rolname,"
+                 "%(columns)s "
+                 "FROM pg_catalog.pg_subscription s "
                  "JOIN pg_catalog.pg_database d "
                  "ON s.subdbid = d.oid "
                  "JOIN pg_catalog.pg_roles AS r "
                  "ON s.subowner = r.oid "
-                 "WHERE s.subname = %(name)s AND d.datname = %(db)s")
+                 "WHERE s.subname = '%(name)s' AND d.datname = '%(db)s'" % {'columns': columns, 'name': self.name, 'db': self.db})
 
         result = exec_sql(self, query, query_params={'name': self.name, 'db': self.db}, add_to_executed=False)
         if result:
@@ -732,13 +739,20 @@ def main():
                                           check_mode=module.check_mode)
 
         else:
-            if connparams:
-                connparams = cast_connparams(connparams)
+            if hasattr(subscription,'connparams'):
+                if connparams:
+                    connparams = cast_connparams(connparams)
 
-            changed = subscription.update(connparams,
-                                          publications,
-                                          subsparams,
-                                          check_mode=module.check_mode)
+                changed = subscription.update(connparams,
+                                            publications,
+                                            subsparams,
+                                            check_mode=module.check_mode)
+            else:
+                module.warn("'connparams' is ignored when existing subscription connparms are unknowable")
+                changed = subscription.update(False,
+                                            publications,
+                                            subsparams,
+                                            check_mode=module.check_mode)
 
         if owner and subscription.attrs['owner'] != owner:
             changed = subscription.set_owner(owner, check_mode=module.check_mode) or changed
--- a/plugins/modules/postgresql_subscription.py
+++ b/plugins/modules/postgresql_subscription.py
@@ -62,6 +62,7 @@ options:
     - The connection dict param-value to connect to the publisher.
     - For more information see U(https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING).
     - Ignored when I(state) is not C(present).
+    - Ignored when an existing subscription's connection parameters are not available from the server (such as in CloudSQL).
     type: dict
   cascade:
     description:
@@ -748,7 +749,7 @@ def main():
                                               subsparams,
                                               check_mode=module.check_mode)
             else:
-                module.warn("'connparams' is ignored when existing subscription connparms are unknowable")
+                module.warn("'connparams' is ignored because pg_subscription.subconninfo is not accessible in your instance.")
                 changed = subscription.update(False,
                                               publications,
                                               subsparams,
--- a/plugins/modules/postgresql_subscription.py
+++ b/plugins/modules/postgresql_subscription.py
@@ -595,14 +595,13 @@ class PgSubscription():
         columns_result = exec_sql(self, columns_sub_table, query_params={'name': self.name, 'db': self.db}, add_to_executed=False)
         columns = ", ".join(["s.%s" % column['column_name'] for column in columns_result])
         query = ("SELECT obj_description(s.oid, 'pg_subscription') AS comment, "
-                 "d.datname, r.rolname,"
-                 "%(columns)s "
+                 "d.datname, r.rolname," + columns + " "
                  "FROM pg_catalog.pg_subscription s "
                  "JOIN pg_catalog.pg_database d "
                  "ON s.subdbid = d.oid "
                  "JOIN pg_catalog.pg_roles AS r "
                  "ON s.subowner = r.oid "
-                 "WHERE s.subname = '%(name)s' AND d.datname = '%(db)s'" % {'columns': columns, 'name': self.name, 'db': self.db})
+                 "WHERE s.subname = %(name)s AND d.datname = %(db)s")
 
         result = exec_sql(self, query, query_params={'name': self.name, 'db': self.db}, add_to_executed=False)
         if result:
--- a/plugins/modules/postgresql_subscription.py
+++ b/plugins/modules/postgresql_subscription.py
@@ -739,7 +739,7 @@ def main():
                                           check_mode=module.check_mode)
 
         else:
-            if hasattr(subscription, 'connparams'):
+            if subscription.attrs['conninfo'] != {}:
                 if connparams:
                     connparams = cast_connparams(connparams)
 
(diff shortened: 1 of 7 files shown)


plugins/modules/postgresql_subscription.py
```diff
@@ -732,13 +739,20 @@ def main():
...
+            if hasattr(subscription, 'connparams'):
+                if connparams:
+                    connparams = cast_connparams(connparams)
+
+                changed = subscription.update(connparams,
+                                              publications,
+                                              subsparams,
+                                              check_mode=module.check_mode)
+            else:
+                module.warn("'connparams' is ignored when existing subscription connparms are unknowable")
```
nickgsc, line 751: This wording is a bit weird. Open to suggestions here on a different way to phrase this so that it is clear why the module won't attempt to make a change to the `connparams`

plugins/modules/postgresql_subscription.py
```diff
@@ -586,15 +586,22 @@ def __get_general_subscr_info(self):
...
-                 "s.subpublications FROM pg_catalog.pg_subscription s "
+                 "d.datname, r.rolname,"
+                 "%(columns)s "
+                 "FROM pg_catalog.pg_subscription s "
                  "JOIN pg_catalog.pg_database d "
                  "ON s.subdbid = d.oid "
                  "JOIN pg_catalog.pg_roles AS r "
                  "ON s.subowner = r.oid "
-                 "WHERE s.subname = %(name)s AND d.datname = %(db)s")
+                 "WHERE s.subname = '%(name)s' AND d.datname = '%(db)s'" % {'columns': columns, 'name': self.name, 'db': self.db})
```
nickgsc, line 604: My python is admittedly pretty weak. For some reason doing the string interpolation this way feels wrong to me, but maybe it's fine idk

hunleyd:
I'm going to hold off on reviewing until @Andersson007 has an opportunity to weigh in

plugins/modules/postgresql_subscription.py
Andersson007: SGTM, Let's discuss it in a separate issue, thanks for spotting this!

plugins/modules/postgresql_subscription.py
Andersson007: - We are passing `name` and `db` values as paramenters to the execute method a bit below (which is a right way to do it) and we don't need to use the single quotes here AFAIK - it should be handled by the method internally.
- We should somehow substitute only columns here. We could manipulate with string conjugation (if i call it correctly) here to do it.

Andersson007 (commented):
@nickgsc thanks for the PR!
- Could you please take a look at my comment.
- Add a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment)

plugins/modules/postgresql_subscription.py
```diff
@@ -732,13 +739,20 @@ def main():
...
+            if subscription.attrs['conninfo'] != {}:
+                if connparams:
+                    connparams = cast_connparams(connparams)
+
+                changed = subscription.update(connparams,
+                                              publications,
+                                              subsparams,
+                                              check_mode=module.check_mode)
+            else:
+                module.warn("'connparams' is ignored when existing subscription connparms are unknowable")
```
Andersson007, line 751:
```suggestion
                module.warn("'connparams' is ignored because it's not supported by the database engine.")
```
Something like that?

nickgsc: That's better, but this could be confusing about it being an issue specific to the implementation/hosting of Postgres, rather than an issue inherent to Postgres itself.

Perhaps we could just say what the actual issue is? Something like

`'connparams' is ignored because pg_subscription.subconninfo is not accessible in your instance.`

Either way, we will want to be very explicit about it in the documentation that will correspond with this. I'll work on that and the changelog fragment today.

plugins/modules/postgresql_subscription.py
nickgsc: >We should somehow substitute only columns here. 

Agreed. That's what I tried to do initially, but wasn't able to make it work because it was mixing different types of string formatting. I'll play with it a bit more

Andersson007 (approved):

Andersson007:
@hunleyd the PR lgtm, any suggestions/concerns?

hunleyd (approved):

Andersson007:
@nickgsc thanks for the contribution! @hunleyd thanks for reviewing!

@nickgsc please check out our [communication guide](https://github.com/ansible-collections/community.postgresql#communication) if you haven't already done that to see other ways how to get in touch with the community

Andersson007:
FYI The [community.postgresql](https://galaxy.ansible.com/ui/repo/published/community/postgresql/) collection version [3.6.0](https://github.com/ansible-collections/community.postgresql/blob/main/CHANGELOG.rst#v360) has been released! Among other changes, it contains fixes for some modules to work with PostgreSQL 17 beta. Thanks to everyone involved!

PR #740, opened 2024-09-27
CI: add testing on Ubuntu 2404, remove testing on Fedora 38

Andersson007:
##### SUMMARY
CI: add testing on Ubuntu 2404, remove testing on Fedora 38


Andersson007:
hm, it's very strange, it fails on both 24 and 22 to restore from a dumb saying that pg_restore fails on executing `SET transaction_timeout = 0`. It's indeed not supported until PG 17.. i also checked it by running it against our `main` - same result..

cameronmurdoch:
I looked at one of the pipeline logs for ubuntu 24.04 and it is seems to be installing postgres 16 server but 17 client.

For task
```
TASK [setup_postgresql_db : Install required OS packages] 
```

Andersson007:
@cameronmurdoch yes, i've just discovered it toon and pushed the fix

Andersson007:
Now it should be fine, i made Ubuntu 2204 to run postgresql 15 as it was supposed to be

Andersson007:
ready for review

hunleyd (changes requested):
did we want to add pg17 since it was released yesterday?

hunleyd (approved):
nevermind, i just saw the issue you opened about pg17 :)

Andersson007:
@hunleyd @cameronmurdoch thanks for reviewing!

PR #741, opened 2024-09-27
Fix units

Andersson007:
##### SUMMARY
Fix units


Andersson007:
something python version dependand that pops up from time to time, ready for review

PR #742, opened 2024-09-27
postgresql_db: fails on setting transaction_timeout = 0 which is supported since PG 17

Andersson007:
##### SUMMARY
Test fails because of this in https://github.com/ansible-collections/community.postgresql/pull/740


Andersson007:
Found the reason, will fix it in https://github.com/ansible-collections/community.postgresql/pull/740, closing this one

Issue #688, opened 2024-04-18
pg_user doesn't take password_encryption into account when checking if a password should be updated


dirkcuys:
##### SUMMARY
When the current password is stored as a md5 hash, but `PGOPTIONS: "-c password_encryption=scram-sha-256"` is passed, the new password will still be hashed with md5. 

https://github.com/ansible-collections/community.postgresql/blob/0bc4754d88a4343a19b97e76e022e9a93fc1fdef/plugins/modules/postgresql_user.py#L473

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
postgresql_user

##### ANSIBLE VERSION

```
ansible [core 2.16.5]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/ZZZ/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /home/ZZZ/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0] (/usr/bin/python3)
  jinja version = 3.0.3
  libyaml = True

```

##### COLLECTION VERSION

```
# /usr/lib/python3/dist-packages/ansible_collections
Collection           Version
-------------------- -------
community.postgresql 3.4.0  
```

##### STEPS TO REPRODUCE

Create a user
```yaml
- name: create postgres user
  postgresql_user:
    name: user
    password: password
    login_host: "127.0.0.1"
```

Update the playbook to use SCRAM

```yaml
- name: create postgres user
  postgresql_user:
    name: user
    password: password
    login_host: "127.0.0.1"
  environment:
    PGOPTIONS: "-c password_encryption=scram-sha-256"
```
And re-run the playbook. The password is reported as unchanged.

##### EXPECTED RESULTS

The password is updated using scram-sha-256

##### ACTUAL RESULTS
The password is unchanged

betanummeric:
Hi @dirkcuys 
Although the `password_encryption` variable may be changed, the password remains the same. Currently the module does not attempt to change the hashing algorithm while keeping the same password.
If you change the password, the module will use the then-specified `password_encryption`.

dirkcuys:
[...]

PR #764, opened 2024-11-08
pg_user doesn't take password_encryption into account when checking if a password should be updated

Andersson007:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.postgresql/issues/688

IMPORTANT:
- The tests I added passed w/o any code change, so I failed to reproduce the bug
- But the tests check that the password is actually stored as MD5
- I carefully debugged and the tests poke the code I added, the code determines correctly that there's `scram-sha-256` on the server by default and executes pwchage = True
- If you think it's risky to add this code, say it, if we agree that it is, I'll close the PR

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/plugins/modules/postgresql_user.py
+++ b/plugins/modules/postgresql_user.py
@@ -185,7 +185,8 @@ notes:
 - SCRAM-SHA-256-hashed passwords (SASL Authentication) require PostgreSQL version 10 or newer.
   On the previous versions the whole hashed string is used as a password.
 - 'Working with SCRAM-SHA-256-hashed passwords, be sure you use the I(environment:) variable
-  C(PGOPTIONS: "-c password_encryption=scram-sha-256") (see the provided example).'
+  C(PGOPTIONS: "-c password_encryption=scram-sha-256") when it is not default
+  for your PostgreSQL version (see the provided example).'
 - On some systems (such as AWS RDS), C(pg_authid) is not accessible, thus, the module cannot compare
   the current and desired C(password). In this case, the module assumes that the passwords are
   different and changes it reporting that the state has been changed.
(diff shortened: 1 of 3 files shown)


Andersson007:
Folks, read the description carefully

@dirkcuys @jogoossens as i failed to reproduce the bug in CI, if you could check it locally if it solves your issue, it'd be great, maybe https://docs.ansible.com/ansible/devel/community/collection_contributors/collection_test_pr_locally.html can help

hunleyd (approved):
this looks OK to me, so approving. holding off on merging in the hopes that @dirkcuys or @jogoossens can test their local scenarios

jogoossens:
Looks good to me to, I hope the patch gets included!

(Testing is not so easy I think, we use execution envs in AWX, would be quite some work :) )

plugins/modules/postgresql_user.py
```diff
@@ -502,21 +508,36 @@ def user_should_we_change_password(current_role_attrs, user, password, encrypted
...
+                hashed_password = 'md5{0}'.format(md5(to_bytes(password) + to_bytes(user)).hexdigest())
+                if hashed_password != current_password:
+                    pwchanging = True
+            elif default_pw_encryption == 'scram-sha-256':
+                # https://github.com/ansible-collections/community.postgresql/issues/688
+                # When the current password is not none and is not
+                # hashed as scram-sha-256 / not explicitly declared as plain text
+                # (if we are here, these conditions should be met)
+                # but the default password encryption is scram-sha-256, update the password.
+                # Can be relevant when migrating from older version of postgres.
```
dirkcuys, line 529: Wouldn't this imply that the password will always be updated if the encryption method is 'scram-sha-256'?

plugins/modules/postgresql_user.py
```diff
@@ -502,21 +508,36 @@ def user_should_we_change_password(current_role_attrs, user, password, encrypted
...
                 pwchanging = True
```
dirkcuys, line 530: Wouldn't this imply that the password will always be updated if the encryption method is 'scram-sha-256'?

Andersson007:
@dirkcuys i saw your comment. Did you deleted it because you saw the condition in line 482 (because I think that case should be handled there) or because of another reason?

dirkcuys:
Sorry about that. I made the comment, then decided to double check by reading through the whole file, not just the diff and as you point out, it's handled on line 482 :)

Andersson007:
> Sorry about that. I made the comment, then decided to double check by reading through the whole file, not just the diff and as you point out, it's handled on line 482 :)

@dirkcuys sure, no problem, thought maybe it's because of another reason, just to re-check. thanks for reviewing!

PR #772, opened 2024-11-15
pg_hba module refactoring

toydarian:
##### SUMMARY

This is a refactoring-only PR and another step on the refactoring and improvement-journey of that module.  
There is a minor change which is, that internally auth-options are now represented with dicts and not with strings anymore. This allows us to verify that the same key doesn't appear multiple times. The integration-tests have been changed to account for that.

There is one more change in the tests, which is due to ordering of rules. I don't know how or why this was the way it was. I'm going with how `ipaddress` sorts those networks, which I think makes sense:
```
In [6]: ipaddress.ip_network('::1/128') < ipaddress.ip_network('2001:db8::1/128')
Out[6]: True
```

Otherwise this doesn't change the modules behavior.

More changes: 
 - adds unit-tests for the refactored rule-class
 - adds comments

In upcoming PRs, I will add
 - handling of `include`
 - taking regexes, groups and inline-includes into account when sorting
 - using standard functions for writes
 - aligning backups with how they are handled elsewhere

##### ISSUE TYPE

- Refactoring Pull Request

##### COMPONENT NAME
postgresql_pg_hba

### DIFF
--- a/changelogs/fragments/772-refactoring-pg_hba-module.yml
+++ b/changelogs/fragments/772-refactoring-pg_hba-module.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - "postgresql_pg_hba - changes ordering of entries that are identical except for the ip-range, but only if the ranges are of the same size, this isn't breaking as ranges of equal size can't overlap (https://github.com/ansible-collections/community.postgresql/pull/772)"
+  - "postgresql_pg_hba - orders auth-options alphabetically, this isn't breaking as the order of those options is not relevant to postgresql (https://github.com/ansible-collections/community.postgresql/pull/772)"
--- a/plugins/modules/postgresql_pg_hba.py
+++ b/plugins/modules/postgresql_pg_hba.py
@@ -934,7 +934,10 @@ class PgHbaRule:
         if self._address:
             ret_dict[header_map['address']] = str(self.source)
         if self._auth_options:
-            ret_dict[header_map['options']] = copy.copy(self._auth_options)
+            # we might want to change the return-type to a dict at some point, but right now we return a string
+            # to not introduce breaking changes
+            # ret_dict[header_map['options']] = copy.copy(self._auth_options)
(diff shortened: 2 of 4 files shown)


hunleyd (approved):
LGTM .. do we need a fragment for refactors @Andersson007 ?

hunleyd:
> Thoughts folks?

i'm ok with either option

Andersson007:
@betanummeric mentioned important points:
- The order in pg_hba file shouldn't change
- I think the order of rows returned by the module are also important, it should be imo the same as in pg_hba, i.e. as it was

@toydarian thoughts?

I would discuss another module for fetching rules like pg_hba_info or something but it's another discussion. Let's make sure now that the module works as it did and merge the PR ASAP

toydarian:
> The order in pg_hba file shouldn't change

The change is insignificant in this case. The order only changes for networks with the same size and it adheres to how the `ipaddress` module orders them. For example, it sorts `'::1/128'` before instead of after `'2001:db8::1/128'`. I can add some code that reverses the order only if the suffix is the same, but I don't think that is justified. It also cannot break anything, as you can't have overlapping networks of the same size.  
The other thing that changes, is that the **options** are now sorted alphabetically, so `ldapserver` comes before `ldapport` before `ldapprefix`, which also doesn't break anything.

> I think the order of rows returned by the module are also important, it should be imo the same as in pg_hba, i.e. as it was

I'll add some code to make it work as before. We can talk about the `info` module later, though I think that would probably not get used a lot.

toydarian:
@Andersson007 @hunleyd changed as discussed, added changelog fragment  
I would have liked to get betanummeric's input, as he gave valuable feedback in the past, but if he doesn't have time, we need to make due without his review

hunleyd (approved):

Andersson007:
@toydarian thanks for the contribution!
@hunleyd @betanummeric thanks for reviewing!

PR #788, opened 2025-01-08
Reconnect to the default database after gathering databases info

kacperk:
##### SUMMARY

`postgresql_info` module fails if the user used for connection doesn't have permission to connect to last database from returned database list.

Fix:
After gathering the information about databases, connection is restored to the database specified in module parameters.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

postgresql_info module

##### ADDITIONAL INFORMATION

Issue with `postgresql_info` module:
When using AWS RDS, you never has access to `rdsadmin` database. If `rdsadmin` databases is return as a last one when listing databases, it keeps trying to connect th `rdsadmin` db when collecting other data, which fails, since you can't connect to the database.

How to reproduce:
Since the database list doesn't have fixed order, it is not that easy to reproduce.

### DIFF
--- a/plugins/modules/postgresql_info.py
+++ b/plugins/modules/postgresql_info.py
@@ -704,7 +704,6 @@ class PgClusterInfo(object):
         self.pg_info["databases"] = db_dict
         self.cursor = self.db_obj.reconnect(default_db)
 
-
     def __get_pretty_val(self, setting):
         """Get setting's value represented by SHOW command."""
         return self.__exec_sql('SHOW "%s"' % setting)[0][setting]
--- a/plugins/modules/postgresql_info.py
+++ b/plugins/modules/postgresql_info.py
@@ -669,6 +669,8 @@ class PgClusterInfo(object):
 
         res = self.__exec_sql(query)
 
+        default_db = self.__exec_sql("SELECT current_database()")[0]["current_database"]
+
         db_dict = {}
         for i in res:
             db_dict[i["datname"]] = dict(
@@ -700,6 +702,8 @@ class PgClusterInfo(object):
                 db_dict[datname]['subscriptions'] = subscr_info.get(datname, {})
 
         self.pg_info["databases"] = db_dict
+        self.cursor = self.db_obj.reconnect(default_db)
+
 
     def __get_pretty_val(self, setting):
         """Get setting's value represented by SHOW command."""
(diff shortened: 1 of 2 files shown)


hunleyd:
`FATAL: Environment `--docker fedora40` is unknown.` any idea what's up with that @Andersson007 ? doesn't seem related to this PR

Andersson007:
> `FATAL: Environment `--docker fedora40` is unknown.` any idea what's up with that @Andersson007 ? doesn't seem related to this PR

@hunleyd it's unrelated to this PR, so feel free to ignore. I've opened another PR to address the failures https://github.com/ansible-collections/community.postgresql/pull/789

plugins/modules/postgresql_info.py
```diff
@@ -700,6 +702,7 @@ def get_db_info(self):
                 db_dict[datname]['subscriptions'] = subscr_info.get(datname, {})
 
         self.pg_info["databases"] = db_dict
+        self.cursor = self.db_obj.reconnect(default_db)
```
Andersson007, line 705:
```suggestion
        # Reconnect to the default DB after gathering info in other DBs
        self.cursor = self.db_obj.reconnect(self.default_db)
```
See the comment above

Andersson007 (changes requested):
@kacperk hi, thanks for the PR! a couple of things from me

plugins/modules/postgresql_info.py
```diff
@@ -230,6 +230,7 @@ def __init__(self, module, db_conn_obj):
         self.module = module
         self.db_obj = db_conn_obj
         self.cursor = db_conn_obj.connect()
+        self.default_db = module.params['db'] or module.params['login_db'] or module.params['database']
```
Andersson007, line 233:
```suggestion
        self.default_db = module.params['db'] or module.params['login_db'] or module.params['database']
```
doesn't it work with just `module.params['login_db']`? the others are just aliases IIRC

kacperk: Im not sure. It seems to work - I put it since just above we override all three values separately, and the conn_params method treat them separately.

hunleyd:
closing/re-opening to tickle the tests

Andersson007 (approved):

plugins/modules/postgresql_info.py
Andersson007: ach, i see, i think it's something specific to reconnect(), let's try w/o it here but keeping an eye after your fix gets released, thanks!

Andersson007 (approved):

hunleyd (approved):

Andersson007:
@kacperk thanks for the contribution! @hunleyd thanks for reviewing and merging!

Issue #790, opened 2025-01-13
PostgreSQL info module fails with Cannot execute SQL syntax error at or near "b'oid'"


chriscroome:
##### SUMMARY

On a new Debian 12 server the info module is failing for me with:

```
Cannot execute SQL 'SELECT s.b'oid', s.b'subdbid', s.b'subskiplsn', s.b'subname', s.b'subowner', s.b'subenabled', s.b'subbinary', s.b'substream', s.b'subtwophasestate', s.b'subdisableonerr', s.b'subconninfo', s.b'subslotname', s.b'subsynccommit', s.b'subpublications', r.rolname AS ownername, d.datname AS dbname FROM pg_catalog.pg_subscription s JOIN pg_catalog.pg_database d ON s.subdbid = d.oid JOIN pg_catalog.pg_roles AS r ON s.subowner = r.oid': syntax error at or near "b'oid'"
LINE 1: SELECT s.b'oid', s.b'subdbid', s.b'subskiplsn', s.b'subname'...
```

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

[community.postgresql.postgresql_info module](https://docs.ansible.com/ansible/latest/collections/community/postgresql/postgresql_info_module.html)

##### ANSIBLE VERSION

```
ansible [core 2.17.7]
  config file = /home/chris/animorph-coop/servers/ansible.cfg
  configured module search path = ['/home/chris/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/chris/.local/pipx/venvs/ansible/lib/python3.11/site-packages/ansible
  ansible collection location = /home/chris/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/chris/.local/bin/ansible
  python version = 3.11.2 (main, Nov 30 2024, 21:22:50) [GCC 12.2.0] (/home/chris/.local/pipx/venvs/ansible/bin/python)
  jinja version = 3.1.5
  libyaml = True
```

##### COLLECTION VERSION

```
# /home/chris/.ansible/collections/ansible_collections
Collection           Version
-------------------- -------
community.postgresql 3.9.1  

# /home/chris/.local/pipx/venvs/ansible/lib/python3.11/site-packages/ansible_collections
Collection           Version
-------------------- -------
community.postgresql 3.9.0  
```

##### CONFIGURATION

```
[...]

PR #791, opened 2025-01-13
postgresql_info: fix SQL error of issue #790

betanummeric:
fixing #790 

The dynamic query construction fails because the column type is `bytes` in python, but the code expects `str`. The python conversion from `bytes` to `str` addess extra characters which break the SQL syntax of the following query.
I guess psycopg 2 vs 3 behave different regarding types, which was unexpected in the code.

This fix using the `quote_ident` function converts the type in SQL which should result in `str` in python. By the way it adds support for arbitrary column names (although this is rather unlikely to become relevant in a table of `pg_catalog`).

```sql
postgres=# SELECT pg_typeof(quote_ident(column_name)), pg_typeof(column_name) FROM information_schema.columns WHERE table_schema = 'pg_catalog' AND table_name = 'pg_subscription' limit 1;
 pg_typeof |             pg_typeof             
-----------+-----------------------------------
 text      | information_schema.sql_identifier
```

### DIFF
--- a/plugins/modules/postgresql_info.py
+++ b/plugins/modules/postgresql_info.py
@@ -327,7 +327,7 @@ class PgClusterInfo(object):
 
     def get_subscr_info(self):
         """Get subscription statistics."""
-        columns_sub_table = ("SELECT column_name "
+        columns_sub_table = ("SELECT quote_ident(column_name) "
                              "FROM information_schema.columns "
                              "WHERE table_schema = 'pg_catalog' "
                              "AND table_name = 'pg_subscription'")


betanummeric:
I didn't test the change yet, I can't promise to do that soon. If someone wants to take this PR over from here - feel free.

Andersson007:
> I didn't test the change yet, I can't promise to do that soon. If someone wants to take this PR over from here - feel free.

@betanummeric thanks for the PR! i think it's OK to wait generally:) thanks for being explicit

Andersson007:
@betanummeric i played with it manually, and it doesn't work
If you look at possible solutions, it'd be great:
- https://github.com/ansible-collections/community.postgresql/pull/804
- https://github.com/ansible-collections/community.postgresql/pull/805

Andersson007:
@betanummeric it turned out that the issue is hard to catch. I'll close this PR as i did with related mine. Feel free to re-open if you think it's relevant. Thanks for investigating and submitting it!

Issue #794, opened 2025-01-20
Regression in postgresql_info for 3.10.0


eb4x:
##### SUMMARY
A task using `postgresql_info` gathering `in_recovery`, stopped working in 3.10.0 collection.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
postgresql_info

##### ANSIBLE VERSION

```paste below
ansible [core 2.16.13]
  config file = /home/erikberg/src/dia-ansible/ansible.cfg
  configured module search path = ['/home/erikberg/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/erikberg/.local/lib/python3.13/site-packages/ansible
  ansible collection location = /home/erikberg/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/erikberg/.local/bin/ansible
  python version = 3.13.1 (main, Dec  9 2024, 00:00:00) [GCC 14.2.1 20240912 (Red Hat 14.2.1-3)] (/usr/bin/python3)
  jinja version = 3.1.4
  libyaml = True

```

##### COLLECTION VERSION

```paste below
Collection           Version
-------------------- -------
community.postgresql 3.10.0 
```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT

EL9 (almalinux)

##### STEPS TO REPRODUCE

```yaml
  tasks:
    - name: Get in_recovery on postgresql
      become: true
      become_user: postgres
      community.postgresql.postgresql_info:
        filter: in_recovery
      register: _pgsql_info
```

##### EXPECTED RESULTS
It just works, like in 3.9.0

##### ACTUAL RESULTS

```paste below
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: KeyError: 'login_db'
[...]

PR #800, opened 2025-01-28
postgresql_info: deprecated db alias in favor of login_db

Andersson007:
##### SUMMARY
postgresql_info: deprecated db alias in favor of login_db

as multiple aliases can actually cause bugs like https://github.com/ansible-collections/community.postgresql/issues/794
i suggest deprecating and removing the `db` one in favor of `login_db`.
`db` is inconsistent with the other login arguments like `login_user`, `login_password`, etc. and it's even misleading. That's why the `login_db` alias was introduced.
I suggest deprecating it now and removing it in the next major release.

Created an [issue](https://github.com/ansible-collections/community.postgresql/issues/801) for milestone 4.0.0.

### DIFF
--- a/changelogs/fragments/0-postgresql_info.yml
+++ b/changelogs/fragments/0-postgresql_info.yml
@@ -1,0 +1,2 @@
+breaking_changes:
+- postgresql_info - the ``db`` alias is deprecated and will be removed in the next major release, use the ``login_db`` argument instead.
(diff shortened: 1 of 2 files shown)


Andersson007:
ready for review
if the majority disagree, we could close it and the issue, no problem

andreasscherbaum:
I'm in favor of this change. Makes the arguments more consistent.

hunleyd (approved):

Andersson007:
@andreasscherbaum @hunleyd thanks for reviewing!

PR #804, opened 2025-01-29
[DRAFT] postgresql_info: fix failure when col names are returned as binary

Andersson007:
##### SUMMARY
One of options to fix https://github.com/ansible-collections/community.postgresql/issues/790 in case it relates to client encoding
Needs manual testing


Andersson007:
as we can't catch the error, closing this for now. in case someone else reports this, we could ask them to test

PR #805, opened 2025-01-29
[DRAFT] postgresql_info: handle encoding

Andersson007:
##### SUMMARY
Relates to https://github.com/ansible-collections/community.postgresql/issues/790
In case it's about server encoding


Andersson007:
as we can't catch the error, closing this for now. in case someone else reports this, we could ask them to test

PR #806, opened 2025-01-31
allow non-string settings in user-specific configuration

### LINKED ISSUE #802: postgresql_user: unable to set desired value for key `search_path` when using `configuration` parameter
##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
postgresql_user

##### ANSIBLE VERSION
2.16.14

##### COLLECTION VERSION

```paste below
ansible.posix        2.0.0  
community.general    10.3.0 
community.postgresql 3.9.1  
```

##### CONFIGURATION

```paste below
not available at the moment
```

##### OS / ENVIRONMENT

Ubuntu 24.04 (running in CI/CD)

##### STEPS TO REPRODUCE

```yaml
    - name: Ensure postgres_exporter DB has correct configuration
      become: true
      become_user: postgres
        name: postgres_exporter
        configuration:
          search_path: postgres_exporter,pg_catalog,public
```

##### EXPECTED RESULTS

Same result as running manually `ALTER ROLE postgres_exporter SET SEARCH_PATH TO postgres_exporter,pg_catalog,public;`, i.e. no quotes around `postgres_exporter, pg_catalog, public`:

```
postgres=# select * from pg_user;
      usename      | usesysid | usecreatedb | usesuper | userepl | usebypassrls |  passwd  | valuntil |                       useconfig
-------------------+----------+-------------+----------+---------+--------------+----------+----------+-------------------------------------------------------
 postgres          |       10 | t           | t        | t       | t            | ******** |          |
 postgres_exporter |    17532 | f           | f        | f       | f            | ******** |          | {"search_path=postgres_exporter, pg_catalog, public"}
```

##### ACTUAL RESULTS

Quotes are inserted around `postgres_exporter, pg_catalog, public`:
```
postgres=# select * from pg_user;
      usename      | usesysid | usecreatedb | usesuper | userepl | usebypassrls |  passwd  | valuntil |                        useconfig
-------------------+----------+-------------+----------+---------+--------------+----------+----------+---------------------------------------------------------
 postgres          |       10 | t           | t        | t       | t            | ******** |          |
 postgres_exporter |    17532 | f           | f        | f       | f            | ******** |          | {"search_path=\"postgres_exporter,pg_catalog,public\""}
```

-------------

With the current behaviour of ansible (and also if I manually run with the value quoted, i.e. `ALTER ROLE postgres_exporter SET SEARCH_PATH TO "postgres_exporter,pg_catalog,public";`), postgres_exporter is unable to use the DB:
`2025-01-28 07:26:40.993 UTC [15624] ERROR:  relation "pg_stat_statements" does not exist at character 393`

However when it is not quoted (`ALTER ROLE postgres_exporter SET SEARCH_PATH TO postgres_exporter,pg_catalog,public;`), the problem is solved.

The module docs state:

> "Inputs to [user](https://docs.ansible.com/ansible/latest/collections/community/postgresql/postgresql_user_module.html#ansible-collections-community-postgresql-postgresql-user-module-parameter-user) as well as keys and values in this parameter are quoted by the module. If you require the user to contain a ", you need to double it, otherwise the module will fail. " and ' are not allowed in configuration keys and any ' in the value of a configuration will be escaped by this module."

So it does not seem like there is a way to avoid the quotes. If this will not be changed then perhaps a warning should be added to the docs because the existence of quotes clearly affects postgresql, at least for `search_path`.

toydarian:
##### SUMMARY

Adds a flag to turn off string-quoting for user-specific configuration, so settings like `search_path` can be set using the module.

Fixes #802 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
postgresql_user

##### ADDITIONAL INFORMATION

This is not optimal, but the best solution I was able to come up with. See discussion in #802

### DIFF
--- a/plugins/modules/postgresql_user.py
+++ b/plugins/modules/postgresql_user.py
@@ -177,6 +177,17 @@ options:
     type: bool
     default: false
     version_added: '3.5.0'
+  quote_configuration_values:
+    description:
+      - Automatically quote the values of configuration variables added via I(configuration). The default is C(true)
+        and setting this to C(false) leaves these options open to SQL-injections and makes the user responsible for
+        properly quoting values.
+      - This is required to be C(false) to modify settings like C(search_path), that need to be unquoted.
+      - If this is C(false) you will also need to make sure that strings are properly quoted.
+        For example C("'16MB'") for C(work_mem).
+      - Set this only to C(false) if you know what you are doing!
+    type: bool
+    default: true
 notes:
 - The module creates a user (role) with login privilege by default.
   Use C(NOLOGIN) I(role_attr_flags) to change this behaviour.
@@ -315,6 +326,14 @@ EXAMPLES = r'''
     configuration:
       work_mem: "16MB"
     reset_unspecified_configuration: true
+
+# Set the search_path
+- name: Set search_path for user
+  community.postgresql.postgresql_user:
+    name: postgres_exporter
+    quote_configuration_values: false
+    configuration:
+      search_path: postgres_exporter, pg_catalog, public
 '''
 
 RETURN = r'''
@@ -1038,7 +1057,7 @@ def parse_user_configuration(module, configs):
         return {}
 
 
-def user_configuration(cursor, module, user, configuration, reset_unspec_config):
+def user_configuration(cursor, module, user, configuration, reset_unspec_config, quote_values):
     """Updates the user's configuration parameters if necessary."""
     current_config_query = "SELECT rolconfig FROM pg_roles WHERE rolname = %(user)s;"
     cursor.execute(current_config_query, {"user": user})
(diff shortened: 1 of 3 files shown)


plugins/modules/postgresql_user.py
```diff
@@ -177,6 +177,17 @@
...
+    description:
+      - Automatically quote the values of configuration variables added via I(configuration). The default is C(true)
+        and setting this to C(false) leaves these options open to SQL-injections and makes the user responsible for
+        properly quoting values.
+      - This is required to be C(false) to modify settings like C(search_path), that need to be unquoted.
+      - If this is C(false) you will also need to make sure that strings are properly quoted.
+        For example C("'16MB'") for C(work_mem).
+      - Set this only to C(false) if you know what you are doing!
+    type: bool
+    default: true
```
Andersson007, line 190:
```suggestion
    default: true
    version_added: '3.11.0'
```

plugins/modules/postgresql_user.py
```diff
@@ -315,6 +326,14 @@
     configuration:
       work_mem: "16MB"
     reset_unspecified_configuration: true
+
+# Set the search_path
```
Andersson007, line 330:
```suggestion
```

Andersson007 (commented):
@toydarian thanks for the PR! A couple of minor things from me

Andersson007:
cc @hunleyd

hunleyd (approved):

Andersson007:
@toydarian thanks for the contribution! @hunleyd thanks for reviewing!
I think we could release as soon as the `db` option is deprecated.

PR #813, opened 2025-03-06
postgresql_publication: add row filter support

gspanos:
##### SUMMARY
Add **row filter** support to `postgresql_publication` module, via an extra yaml parameter named `rowfilters`
This feature is supported since PG Version 15 (search "WHERE" or "row filter"):

- https://www.postgresql.org/docs/15/sql-createpublication.html
- https://www.postgresql.org/docs/15/sql-alterpublication.html

Some design decisions:

1. the parameter allows the user to include or omit the `WHERE` clause. Internally, if the clause exists we remove it and then add it before constructing the SQL fragment. This simplifies, among other things, the row filter comparison described below. 
2. When it comes to idempotency, and detecting whether the row filter has changed or not, I opted for removing any quoting and whitespace and then compare the strings. The reason for that lies in how PG handles quoting and case sensitivity in general. For example if one writes the row filter condition `WHERE ("id" > 10 )`, the double quote(") character won't be retained in the `pg_publication_tables.rowfilter` column. So, short of writing an PG SQL parser, this was the best I could think of.

I have included integration tests for both creation and updating of a publication, and also for testing the "idempotency" handling described above.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
postgresql_publication

##### ADDITIONAL INFORMATION
This will allow a user to declare a row filter, in conjuction with either `tables` or `columns` parameters like so:
```paste below
  - name: Create publication with specific columns and row filter
    postgresql_publication:
      name: test_pub_rowfilter
      columns:
        mytable1: ['id', 'name']
      rowfilters:
        mytable1: '("id" > 10)'
```
or in combination with `tables` parameter
```paste below
  - name: Create publication with specific columns and row filter
    postgresql_publication:
      name: test_pub_rowfilter
      tables: ['mytable1', 'someothertable2']
      rowfilters:
        mytable1: 'WHERE (id > 10)'
```

### DIFF
--- a/plugins/modules/postgresql_publication.py
+++ b/plugins/modules/postgresql_publication.py
@@ -35,6 +35,12 @@ options:
     - Mutually exclusive with I(tables) and I(tables_in_schema).
     type: dict
     version_added: '3.8.0'
+  rowfilters:
+    description:
+    - Optional dictionary of row filters to apply to I(tables) or I(columns) of the publication.
+    - Mutually exclusive with I(tables_in_schema).
+    type: dict
+    version_added: '3.12.0'
   tables:
     description:
     - List of tables to add to the publication.
@@ -120,6 +126,7 @@ seealso:
 author:
 - Loic Blot (@nerzhul) <loic.blot@unix-experience.fr>
 - Andrew Klychkov (@Andersson007) <andrew.a.klychkov@gmail.com>
+- George Spanos (@grantanplan) <spanosgeorge@gmail.com>
 extends_documentation_fragment:
 - community.postgresql.postgres
 '''
@@ -138,7 +145,7 @@ EXAMPLES = r'''
     - prices
     - vehicles
 
-- name: Create publication "acme" publishing only prices table and id and named from vehicles tables
+- name: Create publication "acme" publishing only prices table and id and name from vehicles tables
   community.postgresql.postgresql_publication:
     name: acme
     columns:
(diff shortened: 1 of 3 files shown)


plugins/modules/postgresql_publication.py
```diff
@@ -35,6 +35,12 @@
     - Mutually exclusive with I(tables) and I(tables_in_schema).
     type: dict
     version_added: '3.8.0'
+  rowfilters:
+    description:
+    - Optional dictionary of row filters to apply to I(tables) or I(columns) of the publication. 
+    - Mutually exclusive with I(tables_in_schema).
+    type: dict
+    version_added: '????'
```
Andersson007, line 43:
```suggestion
    version_added: '3.12.0'
```

plugins/modules/postgresql_publication.py
```diff
@@ -336,7 +389,8 @@ def __init__(self, module, cursor, name, pg_srv_ver):
             'parameters': {},
             'owner': '',
             'schemas': [],
-            'columns': {}
+            'columns': {},
+            'rowfilters': {}
```
Andersson007, line 393:
```suggestion
            'rowfilters': {},
```

Andersson007 (commented):
@gspanos thanks for the PR!
Could you please:
- Add a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment), use `minor_changes:`
- Add test tasks with `check_mode: true` to the tests
- Fix the red tests. If you click details and go to AZP, you'll see what the problems are

Andersson007 (approved):
@gspanos thanks!

Could other maintainers please take a look?

hunleyd (approved):

hunleyd:
LGTM

Issue #732, opened 2024-09-09
postgresql_set on log_rotation_size reports change with --check mode


eb4x:
##### SUMMARY

The `postgresql_set` reports a change to `log_rotation_size` when running in `--check` mode.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

`community.postgresql.postgresql_set`

##### ANSIBLE VERSION

```paste below
ansible [core 2.16.4]
  config file = /home/username/src/dia-ansible/ansible.cfg
  configured module search path = ['/home/username/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/username/.local/lib/python3.12/site-packages/ansible
  ansible collection location = /home/username/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/username/.local/bin/ansible
  python version = 3.12.5 (main, Aug  7 2024, 00:00:00) [GCC 13.3.1 20240522 (Red Hat 13.3.1-1)] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```paste below
Collection           Version
-------------------- -------
community.postgresql 3.5.0  
```

##### CONFIGURATION

```paste below
CONFIG_FILE() = /home/username/src/dia-ansible/ansible.cfg
DEFAULT_HOST_LIST(/home/username/src/dia-ansible/ansible.cfg) = ['/home/username/src/dia-ansible/inventory']
DEFAULT_VAULT_IDENTITY_LIST(/home/username/src/dia-ansible/ansible.cfg) = ['app1@~/.ansible/vault/app1', 'app2@~/.ansible/vault/app2']
DEFAULT_VAULT_ID_MATCH(/home/username/src/dia-ansible/ansible.cfg) = True
EDITOR(env: EDITOR) = nvim
```

##### OS / ENVIRONMENT
Target: AlmaLinux 8.10 (Cerulean Leopard)

##### STEPS TO REPRODUCE
Run a playbook to set the `log_rotation_size` for your postgresql server, then run the same playbook with `--check` to see if it needs to change anything.

```yaml
---
- hosts: somehost
  tasks:
    - community.postgresql.postgresql_set:
        name: "{{ item.name }}"
        value: "{{ item.value }}"
      become: true
      become_user: postgres
      loop:
        - { name: log_rotation_size, value: 10240 }
```
I.e.
```
[...]

PR #816, opened 2025-03-27
postgresql_alter_system: new module

Andersson007:
##### SUMMARY
postgresql_alter_system: new module

**Why?**

- postgresql_set was developed by me when i was at the beginning of my contributor journey with all implications
- postgresql_set was developed when i was under pressure at my `$day_job` and family stuff at that time, so with no long term reliable design in mind, i.e. following the "waterfall" , "spagetti code" style
- postgresql_set, considering the above, is buggy and hard to fix w/o breaking. See as example https://github.com/ansible-collections/community.postgresql/issues/732 (i researched)

postgresql_alter_system:
- may be a replacement for postgresql_set, with proper deprecation and enough time for users to switch before removal
- all value type specific logic is isolated in corresponding classes
- the code is generally much better organized imo
- uses the `diff` ret val for mutable attributes and `attrs` ret val for immutable, see what it returns:
```
    "attrs": {
        "boot_val": "try",
        "context": "sighup",
        "enumvals": [
            "off",
            "on",
            "try"
        ],
        "max_val": null,
        "min_val": null,
        "unit": null,
        "vartype": "enum"
    },
    "changed": true,
    "diff": {
        "after": {
            "pending_restart": false,
            "setting": "try"
        },
        "before": {
            "pending_restart": false,
            "setting": "try"
        }
    },
    "executed_queries": [
        "ALTER SYSTEM SET recovery_prefetch = 'False'",
        "SELECT pg_reload_conf()"
    ],
    "restart_required": false
```
- the interface is clearer (no other module-specific options)
```
- name: Set work_mem
  community.postgresql.postgresql_alter_system:
     param: work_mem
     value: 1024

- name: Reset work_mem
  community.postgresql.postgresql_alter_system:
     param: work_mem
     value: _RESET
```

**Questions/considerations**

- If you support the idea, I'll add unit tests
- What do you folks think?
- Should we expose the `restart_required` ret value or it's fine to check it using `diff`?

### DIFF
--- a/plugins/modules/postgresql_alter_system.py
+++ b/plugins/modules/postgresql_alter_system.py
@@ -266,8 +266,9 @@ class PgParam():
         query = ("SELECT setting, unit, context, vartype, enumvals, "
                  "boot_val, reset_val, pending_restart "
                  "FROM pg_settings where name = %s")
+        executed_queries.append(query % self.name)
         res = self.__exec_sql(query, (self.name,))
-        # DEBUG
+        executed_queries.append(res[0])  # TODO remove this DEBUG
         return res[0]
 
     def __exec_sql(self, query, params=()):
--- a/plugins/modules/postgresql_alter_system.py
+++ b/plugins/modules/postgresql_alter_system.py
@@ -209,6 +209,14 @@ class ValueBool(Value):
             return value
 
 
+class ValueInt(Value):
+
+    def __init__(self, module, param_name, value, default_unit):
+        self.module = module
+        self.default_unit = None  # TODO Evaluate later if you need it
+        self.normalized = value
+
+
 class ValueMem(Value):
     # If you pass anything else for memory-related param,
     # Postgres will show that only the following
@@ -285,7 +293,8 @@ def build_value_class(module, param_name, value, unit, vartype):
             return ValueMem(module, param_name, value, unit)
         else:
             # TODO change it to a specific case
-            return ValueMem(module, param_name, value, unit)
+            return ValueInt(module, param_name, value, unit)
+
     elif vartype == 'bool':
         return ValueBool(module, param_name, value, unit)
 
--- a/plugins/modules/postgresql_alter_system.py
+++ b/plugins/modules/postgresql_alter_system.py
@@ -70,8 +70,8 @@ seealso:
 - name: PostgreSQL server configuration
   description: General information about PostgreSQL server configuration.
   link: https://www.postgresql.org/docs/current/runtime-config.html
-- name: PostgreSQL view pg_settings reference
-  description: Complete reference of the pg_settings view documentation.
+- name: PostgreSQL view pg_catalog.pg_settings reference
+  description: Complete reference of the pg_catalog.pg_settings view documentation.
   link: https://www.postgresql.org/docs/current/view-pg-settings.html
 - name: PostgreSQL ALTER SYSTEM command reference
   description: Complete reference of the ALTER SYSTEM command documentation.
@@ -104,7 +104,7 @@ EXAMPLES = r'''
 RETURN = r'''
 attrs:
   description:
-  - Parameter attributes from C(pg_settings)
+  - Parameter attributes from C(pg_catalog.pg_settings)
     that do not change.
   returned: success
   type: dict
@@ -121,7 +121,7 @@ diff:
   - A dictionary the C(before) and C(after) keys.
   - Each key contains a dictionary of key-value pairs
     representing changeable columns and values for the parameter
-    obtained from the pg_settings relation.
+    obtained from the pg_catalog.pg_settings relation.
   returned: success
   type: dict
   sample: {
@@ -256,7 +256,7 @@ def check_problematic_params(module, param, value):
 
 class ValueBool():
     """Represents a parameter of type bool."""
-    # SELECT * FROM pg_settings WHERE vartype = 'bool'
+    # SELECT * FROM pg_catalog.pg_settings WHERE vartype = 'bool'
 
     def __init__(self, module, param_name, value, default_unit, pg_ver=None):
         # We do not use all the parameters in every class
@@ -273,7 +273,7 @@ class ValueInt():
     """Represents a parameter of type integer.
     Memory- and time-related parameters are handled by dedicated classes.
     """
-    # SELECT * FROM pg_settings WHERE vartype = 'integer' and unit IS NULL
+    # SELECT * FROM pg_catalog.pg_settings WHERE vartype = 'integer' and unit IS NULL
 
     def __init__(self, module, param_name, value, default_unit, pg_ver=None):
         # We do not use all the parameters in every class
@@ -285,7 +285,7 @@ class ValueInt():
 
 class ValueString():
     """Represents a parameter of type string."""
-    # SELECT * FROM pg_settings WHERE vartype = 'string'
+    # SELECT * FROM pg_catalog.pg_settings WHERE vartype = 'string'
 
     def __init__(self, module, param_name, value, default_unit, pg_ver):
         # We do not use all the parameters in every class
@@ -308,7 +308,7 @@ class ValueString():
 
 class ValueEnum():
     """Represents a parameter of type enum."""
-    # SELECT * FROM pg_settings WHERE vartype = 'enum'
+    # SELECT * FROM pg_catalog.pg_settings WHERE vartype = 'enum'
 
     def __init__(self, module, param_name, value, default_unit, pg_ver=None):
         # We do not use all the parameters in every class
@@ -326,7 +326,7 @@ class ValueEnum():
 def normalize_bool_val(value):
     # No idea why Ansible converts on/off passed as string
     # to "True" and "False". However, there are represented
-    # as "on" and "off" in pg_settings.
+    # as "on" and "off" in pg_catalog.pg_settings.
     if value == "True":
         return "on"
     elif value == "False":
@@ -337,7 +337,7 @@ def normalize_bool_val(value):
 
 class ValueReal():
     """Represents a parameter of type real."""
-    # SELECT * FROM pg_settings WHERE vartype = 'real'
+    # SELECT * FROM pg_catalog.pg_settings WHERE vartype = 'real'
 
     def __init__(self, module, param_name, value, default_unit, pg_ver=None):
         # We do not use all the parameters in every class
@@ -414,7 +414,7 @@ class ValueTime():
 
         # When it doesn't contain a unit part
         # we set it as the unit defined for this
-        # parameter in pg_settings
+        # parameter in pg_catalog.pg_settings
         else:
             int_part = to_int(self.module, value)
             unit_part = self.default_unit
@@ -451,7 +451,7 @@ class ValueMem():
         self.default_unit = default_unit
         self.num_value, self.passed_unit = self.__set(param_name, value)
         if self.passed_unit == "8kB":
-            # This is a special case when the unit in pg_settings is "8kB".
+            # This is a special case when the unit in pg_catalog.pg_settings is "8kB".
             # Users can still pass such values as "10MB", etc.
             # The only issue seems to appear when users don't specify values
             # of 8kB default value explicitly, i.e., when they pass just "100".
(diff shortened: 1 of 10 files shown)


Andersson007:
@hunleyd @toydarian @betanummeric , Andreas and George and others ^

hunleyd:
* If you support the idea, I'll add unit tests

+1 from me. <3 this

* Should we expose the restart_required ret value or it's fine to check it using diff?

I'm OK with either

hunleyd (approved):

YorviArias:
> If you support the idea, I'll add unit tests

This looks like a great idea to me. The interface looks similar to set module so hopefully will be easy for others to move to it.

keithf4:
Sounds great. I think it would be better to expose the pending restart somehow. Be good to be able to easily check that.

plugins/modules/postgresql_alter_system.py
```diff
@@ -0,0 +1,876 @@
...
+    postgres_common_argument_spec,
+)
+
+executed_queries = []
+
+PG_SUPPORTED_VER = 140000
+
+# GUC_LIST_QUOTE parameters list for each version where they changed (from PG_REQ_VER).
+# It is a tuple of tuples as we need to iterate it in order.
+# TODO it was copied here from postgresql_set. After merge, it should be
```
andreasscherbaum, line 177: If `postgresql_set` is sunset, will that not solve the problem?

plugins/modules/postgresql_alter_system.py
```diff
@@ -0,0 +1,876 @@
...
+        # this will always run the command to ensure the removal
+        # and report changed=true
+        query = "ALTER SYSTEM RESET %s" % self.name
+        self.__exec_set_sql(query)
+        return True
+
+    def get_attrs(self):
+        query = ("SELECT setting, unit, context, vartype, enumvals, "
+                 "boot_val, min_val, max_val, pending_restart "
+                 "FROM pg_settings where name = %s")
```
andreasscherbaum, line 661: Please make this `pg_catalog.pg_settings`.

plugins/modules/postgresql_alter_system.py
```diff
@@ -0,0 +1,876 @@
...
+
+    def __check_param_context(self, context):
+        if context == "internal":
+            msg = ("%s cannot be changed (internal context). "
+                   "See https://www.postgresql.org/docs/current/"
+                   "runtime-config-preset.html" % self.name)
+            self.module.fail_json(msg=msg)
+
+        elif context == "postmaster":
+            self.module.warn("Restart of PostgreSQL is required for setting %s" % self.name)
```
andreasscherbaum, line 710: Is it useful to drop a warning to the console/log every time such a value is changed?
How about making this a debug message: if someone doesn't know, this person will run debug and then see the message.

tests/integration/targets/postgresql_alter_system/tasks/test_bools.yml
```diff
@@ -0,0 +1,110 @@
...
+      - result["diff"]["before"]["pending_restart"] == False
+      - result["diff"]["before"]["setting"] == "on"
+      - result["diff"]["after"]["pending_restart"] == False
+      - result["diff"]["after"]["setting"] == "on"
+
+  - name: Check nothing has changed
+    <<: *task_parameters
+    postgresql_query:
+      <<: *pg_parameters
+      query: "SELECT setting FROM pg_settings WHERE name = 'autovacuum'"
```
andreasscherbaum, line 44: `pg_catalog.pg_settings`

tests/integration/targets/postgresql_alter_system/tasks/test_bools.yml
```diff
@@ -0,0 +1,110 @@
...
+      - result["diff"]["before"]["pending_restart"] == False
+      - result["diff"]["before"]["setting"] == "on"
+      - result["diff"]["after"]["pending_restart"] == False
+      - result["diff"]["after"]["setting"] == "off"
+
+  - name: Check it has changed
+    <<: *task_parameters
+    postgresql_query:
+      <<: *pg_parameters
+      query: "SELECT setting FROM pg_settings WHERE name = 'autovacuum'"
```
andreasscherbaum, line 80: `pg_catalog.pg_settings`

tests/integration/targets/postgresql_alter_system/tasks/test_enums.yml
```diff
@@ -0,0 +1,123 @@
...
+      - result["diff"]["before"]["pending_restart"] == False
+      - result["diff"]["before"]["setting"] == "try"
+      - result["diff"]["after"]["pending_restart"] == False
+      - result["diff"]["after"]["setting"] == "try"
+
+  - name: Check it has changed
+    <<: *task_parameters
+    postgresql_query:
+      <<: *pg_parameters
+      query: "SELECT setting FROM pg_settings WHERE name = 'recovery_prefetch'"
```
andreasscherbaum, line 45: `pg_catalog.pg_settings`

tests/integration/targets/postgresql_alter_system/tasks/test_enums.yml
```diff
@@ -0,0 +1,123 @@
...
+      - result["diff"]["before"]["pending_restart"] == False
+      - result["diff"]["before"]["setting"] == "try"
+      - result["diff"]["after"]["pending_restart"] == False
+      - result["diff"]["after"]["setting"] == "off"
+
+  - name: Check it has changed
+    <<: *task_parameters
+    postgresql_query:
+      <<: *pg_parameters
+      query: "SELECT setting FROM pg_settings WHERE name = 'recovery_prefetch'"
```
andreasscherbaum, line 81: `pg_catalog.pg_settings`

tests/integration/targets/postgresql_alter_system/tasks/test_reals.yml
```diff
@@ -0,0 +1,109 @@
...
+      - result["diff"]["before"]["pending_restart"] == False
+      - result["diff"]["before"]["setting"] == 0.1
+      - result["diff"]["after"]["pending_restart"] == False
+      - result["diff"]["after"]["setting"] == 1
+
+  - name: Check it has changed
+    <<: *task_parameters
+    postgresql_query:
+      <<: *pg_parameters
+      query: "SELECT setting FROM pg_settings WHERE name = 'autovacuum_analyze_scale_factor'"
```
andreasscherbaum, line 80: `pg_catalog.pg_settings`

tests/integration/targets/postgresql_alter_system/tasks/test_strings.yml
```diff
@@ -0,0 +1,233 @@
...
+      - result["diff"]["before"]["pending_restart"] == False
+      - result["diff"]["before"]["setting"] != "Europe/Berlin"
+      - result["diff"]["after"]["pending_restart"] == False
+      - result["diff"]["after"]["setting"] != "Europe/Berlin"
+
+  - name: Check nothing has changed
+    <<: *task_parameters
+    postgresql_query:
+      <<: *pg_parameters
+      query: "SELECT setting FROM pg_settings WHERE name = 'log_timezone'"
```
andreasscherbaum, line 58: `pg_catalog.pg_settings`

tests/integration/targets/postgresql_alter_system/tasks/test_strings.yml
```diff
@@ -0,0 +1,233 @@
...
+      - result["diff"]["before"]["pending_restart"] == False
+      - result["diff"]["before"]["setting"] != "Europe/Berlin"
+      - result["diff"]["after"]["pending_restart"] == False
+      - result["diff"]["after"]["setting"] == "Europe/Berlin"
+
+  - name: Check it has changed
+    <<: *task_parameters
+    postgresql_query:
+      <<: *pg_parameters
+      query: "SELECT setting FROM pg_settings WHERE name = 'log_timezone'"
```
andreasscherbaum, line 93: `pg_catalog.pg_settings`

tests/integration/targets/postgresql_alter_system/tasks/test_strings.yml
```diff
@@ -0,0 +1,233 @@
...
+      - result["diff"]["before"]["pending_restart"] == False
+      - result["diff"]["before"]["setting"] == '"$user", public'
+      - result["diff"]["after"]["pending_restart"] == False
+      - result["diff"]["after"]["setting"] == '"$user", public'
+
+  - name: Check nothing has changed
+    <<: *task_parameters
+    postgresql_query:
+      <<: *pg_parameters
+      query: "SELECT setting FROM pg_settings WHERE name = 'search_path'"
```
andreasscherbaum, line 182: `pg_catalog.pg_settings`

Andersson007:
@andreasscherbaum @hunleyd @YorviArias @keithf4 thanks for the feedback!
converted it to draft and will proceed with incorporating it
any more feedback along the way is appreciated)

tests/integration/targets/postgresql_alter_system/tasks/test_bools.yml
Andersson007: nice shot, will change, thanks!

plugins/modules/postgresql_alter_system.py
Andersson007: Could you please elaborate a bit more on this?

Andersson007: Do you mean why are we pulling stuff from postgresql_set?
It wasn't me who write this and I'm not sure how important these function are. I should probably change this todo to add something like re-evaluate if we need this and remove it if not.
As for now, i think it's more reliable to have these two functions here. Anyway, there are only two auxiliary functions from there.
Do i understand the question correctly?

plugins/modules/postgresql_alter_system.py
Andersson007: @andreasscherbaum how about this https://github.com/ansible-collections/community.postgresql/pull/816/commits/fc1f15b9b4d192618dd63d6ffb921dfaa91298be ?

plugins/modules/postgresql_alter_system.py
Andersson007: @andreasscherbaum ah, yes, sorry, that was the point, you're right. if `_set` goes anyway (hopefully), there's no point to move it around. I'll update the comments

plugins/modules/postgresql_alter_system.py
Andersson007: oh, looks like python 2.7 (we still must support) does not support empty dict unpacking..

Andersson007: another version https://github.com/ansible-collections/community.postgresql/pull/816/commits/7bd2e044f2bdd93cc6fb5124b031d1b0ed8d8e30
(the tests from the commit above are still relevant).
PTAL

andreasscherbaum: My "concern" is why you are dropping a warning here, and not a notice or similar.
This will output a warning on the console, somewhere in the flow.

I can see two possible scenarios:

1. The author of the Playbook already knows about the need to restart, in this case the warning is not necessary
2. The author doesn't know, but will fiddle around why the change is not working, and debugging (instead of a warning) will show the message

I can't recollect seeing warnings displayed in other Ansible modules when a config change requires a restart.

Andersson007:
Done with the feedback put here so far. PTAL
Next, will try to add some unit tests. Would take some time, probably next week. FYI
It's still open for more feedback if anybody has anything

andreasscherbaum:
The comment about the warning is still open. Otherwise looks good.

plugins/modules/postgresql_alter_system.py
Andersson007: in the current implementation, there'll be no warning in the output, just a note in the returned json.
do you think it's OK not to have it at all? I'd be OK with it too. We could put a short note in the doc about such parameters and emphasize the `restart_required` return value.
what do you think?

andreasscherbaum: Yes, I think that's ok. When I look at this, I don't see the development phase, but the phase where this playbook runs in production. At this point the playbook is already well-tested and mature enough to already cover the restart.

A note and a debug message is fine in my opinion.

Andersson007:
Folks, can we merge it now?
I'd like to cover those data transforming functions with unit tests in a separate PR before the next release (which we aren't planning to do this week anyway, so should be fine i believe).

andreasscherbaum (approved):
Looks good now.

hunleyd (approved):

Andersson007:
thanks everyone for reviewing!

Issue #820, opened 2025-03-31
Deprecate the `rename` choice from modules. Schedule removal in one of the next major releases


Andersson007:
##### SUMMARY

See the discussion in https://github.com/ansible-collections/community.postgresql/issues/815

Main issues with this choice:
- It's not idempotent
- It's more like an adhoc command
- It's even dangerous as was mentioned by @andreasscherbaum

It's easy to implement it with postgresql_query

Andersson007:
Thoughts / concerns? In the meantime, i'll put the help_wanted label

PR #832, opened 2025-04-13
Deprecate rename and rename_to choices/options

Andersson007:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.postgresql/issues/820

### DIFF
--- a/changelogs/fragments/1-deprecate_rename.yml
+++ b/changelogs/fragments/1-deprecate_rename.yml
@@ -1,5 +1,5 @@
 deprecated_features:
-- postgresql_db - the ``rename`` choice of the state option is deprecated and will be removed in version 5.0.0, use the postgresql_query module instead.
-- postgresql_table - the ``rename`` option is deprecated and will be removed in version 5.0.0, use the postgresql_query module instead.
-- postgresql_tablespace - the ``rename_to`` option is deprecated and will be removed in version 5.0.0, use the postgresql_query module instead.
-- postgresql_sequence - the ``rename_to`` option is deprecated and will be removed in version 5.0.0, use the postgresql_query module instead.
+- postgresql_db - the ``rename`` choice of the state option is deprecated and will be removed in version 5.0.0, use the ``postgresql_query`` module instead.
+- postgresql_table - the ``rename`` option is deprecated and will be removed in version 5.0.0, use the ``postgresql_query module`` instead.
+- postgresql_tablespace - the ``rename_to`` option is deprecated and will be removed in version 5.0.0, use the ``postgresql_query`` module instead.
+- postgresql_sequence - the ``rename_to`` option is deprecated and will be removed in version 5.0.0, use the ``postgresql_query`` module instead.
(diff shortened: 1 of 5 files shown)


andreasscherbaum:
I'd love to see an explanation as to why this is deprecated, but I don't know what is the right place for it.

Basically summarizing the discussion, pointing out that using "rename" is not playing well with state-based systems like Ansible.

changelogs/fragments/1-deprecate_rename.yml
```diff
@@ -0,0 +1,5 @@
+deprecated_features:
+- postgresql_db - the ``rename`` choice of the state option is deprecated and will be removed in version 5.0.0, use the postgresql_query module instead.
+- postgresql_table - the ``rename`` option is deprecated and will be removed in version 5.0.0, use the postgresql_query module instead.
+- postgresql_tablespace - the ``rename_to`` option is deprecated and will be removed in version 5.0.0, use the postgresql_query module instead.
+- postgresql_sequence - the ``rename_to`` option is deprecated and will be removed in version 5.0.0, use the postgresql_query module instead.
```
hunleyd, line 5:
```suggestion
- postgresql_db - the ``rename`` choice of the state option is deprecated and will be removed in version 5.0.0, use the ``postgresql_query`` module instead.
- postgresql_table - the ``rename`` option is deprecated and will be removed in version 5.0.0, use the ``postgresql_query module`` instead.
- postgresql_tablespace - the ``rename_to`` option is deprecated and will be removed in version 5.0.0, use the ``postgresql_query`` module instead.
- postgresql_sequence - the ``rename_to`` option is deprecated and will be removed in version 5.0.0, use the ``postgresql_query`` module instead.
```
thoughts?

hunleyd (changes requested):

Andersson007:
> I'd love to see an explanation as to why this is deprecated, but I don't know what is the right place for it.
> 
> Basically summarizing the discussion, pointing out that using "rename" is not playing well with state-based systems like Ansible.

@andreasscherbaum SGTM, i would avoid putting to much in the option's docs tough, how about:

Option 1: we add a link to the GH discussion which would give the full context, something like:
```
DEPRECATED (see the L(discussion,https://github.com/ansible-collections/community.postgresql/issues/820)). C(rename) is used to rename the database C(name) to C(target).
To rename a database, use the M(community.postgresql.postgresql_query) module.
```
Option 2: Add a more verbose note to the `notes:` section of the `DOCUMENTATION` block and refer from the option's doc to the note (though I personally like the link to the discussion option).

Thoughts, folks?

andreasscherbaum:
Option 1 is already fine, I think. It's about giving users an idea as to why this is deprecated. Otherwise the discussion might start again.

Andersson007:
> Option 1 is already fine, I think. It's about giving users an idea as to why this is deprecated. Otherwise the discussion might start again.

Done, PTAL, thanks

andreasscherbaum (approved):
Looks good.

hunleyd (approved):

Andersson007:
@andreasscherbaum @hunleyd thanks for reviewing and approving the changes!

Issue #834, opened 2025-04-15
Fix CI run against the devel ansible-core branch


Andersson007:
##### SUMMARY

While working on another PR, i've noticed that some new checks had arrived with ansible-core devel branch.
Now the jinja templating is more strict and things like the above are not allowed any more:
```
06:59 [ERROR]: Task failed: Syntax error in expression. Template delimiters are not supported in expressions: expected name or number
06:59 
06:59 Task failed.
06:59 Origin: /root/ansible_collections/community/postgresql/tests/output/.tmp/integration/postgresql_membership-_kobo957-ÅÑŚÌβŁÈ/tests/integration/targets/postgresql_membership/tasks/postgresql_membership_initial.yml:41:3
06:59 
06:59 39   check_mode: true
06:59 40
06:59 41 - assert:
06:59      ^ column 3
06:59 
06:59 <<< caused by >>>
06:59 
06:59 Syntax error in expression. Template delimiters are not supported in expressions: expected name or number
06:59 Origin: /root/ansible_collections/community/postgresql/tests/output/.tmp/integration/postgresql_membership-_kobo957-ÅÑŚÌβŁÈ/tests/integration/targets/postgresql_membership/tasks/postgresql_membership_initial.yml:46:7
06:59 
06:59 44     - result.groups == ["{{ test_group1 }}"]
06:59 45     - result.queries == ["GRANT \"{{ test_group1 }}\" TO \"{{ test_user1 }}\""]
06:59 46     - result.granted.{{ test_group1 }} == ["{{ test_user1 }}"]
06:59          ^ column 7
06:59 
06:59 fatal: [testhost]: FAILED! => {
06:59     "changed": false,
06:59     "msg": "Task failed: Syntax error in expression. Template delimiters are not supported in expressions: expected name or number"
06:59 }
```

We would appreciate your help in fixing this.
What you need to do:

1. Install ansible-core from github repo and clone the collection, see the [quick-start guide](https://docs.ansible.com/ansible/devel/community/create_pr_quick_start.html), in particular:
  - clone https://github.com/ansible/ansible repo locally
  - in the same terminal window where you'll run the sanity tests and being in the ansible source directory run `source hacking/env-setup`
2. go to you collection cloned directly (follow the guide ^) and run the tests
3. fix and run, repeat until they pass

If you need any help/advice, ask

PR #837, opened 2025-04-22
Fix CI

Andersson007:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.postgresql/issues/834

Seems that some fancy jinja changes have been added to the devel branch..
FYI some removed checks are not as important because we check the actual DB state anyway. We shouldn't imo invest too much time in investigating that stuff as long as the CI is green


Andersson007:
ready for review

hunleyd (approved):

PR #846, opened 2025-05-05
Deprecate some login aliases

Andersson007:
##### SUMMARY
Deprecate some login aliases

FYI there's also ``port`` (with ``login_port`` alias), but as it's not an alias I would tackle it in a separate PR later (no rush with it).


Andersson007:
ready for review

andreasscherbaum (approved):

hunleyd (approved):

Issue #853, opened 2025-05-13
[3.14.0] postgresql_alter_system module is trying to int a non-int param


badfiles:
ENV:
ansible [core 2.16.14]
  python version = 3.12.3 (main, Feb  4 2025, 14:48:35) [GCC 13.3.0] (/usr/bin/python3)

OS:
NAME="AlmaLinux"
VERSION="9.5 (Teal Serval)"
ID="almalinux"
ID_LIKE="rhel centos fedora"
VERSION_ID="9.5"

postgres (PostgreSQL) 15.13

```
postgres=# SELECT setting, unit, context, vartype, enumvals,
                 boot_val, min_val, max_val, pending_restart
                 FROM pg_catalog.pg_settings where name = 'random_page_cost'
;
 setting | unit | context | vartype | enumvals | boot_val | min_val |   max_val    | pending_restart 
---------+------+---------+---------+----------+----------+---------+--------------+-----------------
 1.1     |      | user    | real    |          | 4        | 0       | 1.79769e+308 | f
(1 row)

An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ValueError: invalid literal for int() with base 10: '1.79769e+308'
failed: [****] (item={'name': 'random_page_cost', 'value': '1.1'}) => changed=false 
  ansible_loop_var: item
  item:
    name: random_page_cost
    value: '1.1'
  module_stderr: |-
    Traceback (most recent call last):
      File "<stdin>", line 107, in <module>
      File "<stdin>", line 99, in _ansiballz_main
      File "<stdin>", line 47, in invoke_module
      File "/usr/lib64/python3.9/runpy.py", line 225, in run_module
        return _run_module_code(code, init_globals, run_name, mod_spec)
      File "/usr/lib64/python3.9/runpy.py", line 97, in _run_module_code
        _run_code(code, mod_globals, init_globals,
      File "/usr/lib64/python3.9/runpy.py", line 87, in _run_code
        exec(code, run_globals)
      File "/tmp/ansible_community.postgresql.postgresql_alter_system_payload_g5aegz2y/ansible_community.postgresql.postgresql_alter_system_payload.zip/ansible_collections/community/postgresql/plugins/modules/postgresql_alter_system.py", line 874, in <module>
[...]

PR #854, opened 2025-05-14
postgresql_alter_system: fix values returned in scientific notation

Andersson007:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.postgresql/issues/853

### DIFF
--- a/plugins/modules/postgresql_alter_system.py
+++ b/plugins/modules/postgresql_alter_system.py
@@ -572,7 +572,16 @@ def convert_ret_vals(attrs):
         if str_contains_float(attrs[elem]):
             attrs[elem] = float(attrs[elem])
         else:
-            attrs[elem] = int(attrs[elem])
+            try:
+                attrs[elem] = int(attrs[elem])
+            except ValueError:
+                # Leave the attrs[elem] as-is, i.e. a string
+                # This can happen when max_value of type real
+                # is huge and written in scientific notation
+                # e.g., 1.79769e+308 (as of 2025-05-14, this is the only one)
+                # and it doesn't make sense to convert it as it'll be really huge
+                # https://github.com/ansible-collections/community.postgresql/issues/853
+                pass
 
     return attrs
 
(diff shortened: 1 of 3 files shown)


Andersson007:
ready for review

toydarian (approved):

hunleyd (approved):

andreasscherbaum:
I think this needs more work:

```
ads=# set work_mem to '1.79769e+308';
ERROR:  invalid value for parameter "work_mem": "1.79769e+308"
HINT:  Value exceeds integer range.
```

My assumption is that the most common case for value+exponent is that the actual value is too large. And then PG will also fail.

plugins/modules/postgresql_alter_system.py
```diff
@@ -572,7 +572,16 @@ def convert_ret_vals(attrs):
         if str_contains_float(attrs[elem]):
             attrs[elem] = float(attrs[elem])
         else:
-            attrs[elem] = int(attrs[elem])
+            try:
+                attrs[elem] = int(attrs[elem])
+            except ValueError:
```
andreasscherbaum, line 577: If the cast to `int()` fails, it is likely or at least possible that PostgreSQL doesn't like the value either.
It doesn't help if the Ansible code passes, just to fail moments later in the database.

Andersson007: @andreasscherbaum shouldn't we just return what DB returns, because if people run a select against max_val it'll return "1.79769e+308". I suggest leaving it to PG devs and just return what DB returns in similar cases. Thoughts?

andreasscherbaum: You have a point there, and this is just the return value. If someone uses this same value to set a parameter, this will work. If someone uses an arbitrary value, this might fail - but that is a different problem.

andreasscherbaum (approved):
All questions solved, good to merge

Andersson007:
Thanks for reviewing, everyone!

PR #858, opened 2025-05-20
fix: report change in check mode when schema exists and state is absent

allartk:
##### SUMMARY

Removing an existing schema in check mode does not report changes.
However removing a non-existing schema in check mode does.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

postgresql_schema

##### ADDITIONAL INFORMATION

Example

```yaml
- name: Remove londiste schema from database
  become: true
  become_user: postgres
  community.postgresql.postgresql_schema:
    name: londisteee
    login_db: "{{ item.db }}"
    port: "{{ item.port }}"
    state: absent
    cascade_drop: true
  loop: "{{ londiste_cleanup }}"
```

### DIFF
--- a/plugins/modules/postgresql_schema.py
+++ b/plugins/modules/postgresql_schema.py
@@ -288,7 +288,7 @@ def main():
     try:
         if module.check_mode:
             if state == "absent":
-                changed = not schema_exists(cursor, schema)
+                changed = schema_exists(cursor, schema)
             elif state == "present":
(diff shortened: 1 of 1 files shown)


allartk:
PS: could use some help (if the change is correct) with the tests...

allartk:
Thanks for the feedback, I'll try to fix, as soon as my schedule allows it.

hunleyd (approved):

PR #860, opened 2025-05-28
Deprecate port option

Andersson007:
##### SUMMARY
Deprecate port option + fixed some newly arrived test failures

### DIFF
--- a/tests/integration/targets/setup_postgresql_db/tasks/main.yml
+++ b/tests/integration/targets/setup_postgresql_db/tasks/main.yml
@@ -2,7 +2,6 @@
 # WARNING: These are designed specifically for Ansible tests       #
 # and should not be used as examples of how to write Ansible roles #
 ####################################################################
-
 - name: python 2
   set_fact:
     python_suffix: ''
@@ -49,15 +48,14 @@
 
 - name: remove old db config and files (debian)
   file:
-    path: '{{ loop_item }}'
+    path: '{{ item }}'
     state: absent
   ignore_errors: true
   when: ansible_os_family == "Debian"
   loop:
+  - '{{ pg_dir }}'
   - /etc/postgresql
   - /var/lib/postgresql
-  loop_control:
-    loop_var: loop_item
 
 #
 # Prepare Ubuntu for PostgreSQL install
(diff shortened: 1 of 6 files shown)


Andersson007:
ready for review

tests/integration/targets/setup_postgresql_db/tasks/main.yml
```diff
@@ -49,15 +48,14 @@
...
-    path: '{{ loop_item }}'
+    path: '{{ item }}'
     state: absent
   ignore_errors: true
   when: ansible_os_family == "Debian"
   loop:
+  - '{{ pg_dir }}'
   - /etc/postgresql
   - /var/lib/postgresql
-  loop_control:
```
andreasscherbaum, line 59: Out of curiosity: why remove the loop control?

tests/unit/plugins/module_utils/test_postgres.py
```diff
@@ -31,7 +31,7 @@
     user=dict(default='postgres'),
     password=dict(default='test', no_log=True),
     host=dict(default='test'),
-    port=dict(type='int', default=5432, aliases=['login_port']),
+    port=dict(type='int', aliases=['port'], default=5432),
```
andreasscherbaum, line 34: That alias is supposed to be `login_port`.

andreasscherbaum, line 34: Is the variable name supposed to be `port` or `login_port`.

Andersson007: @andreasscherbaum this dict is used in the test_get_conn_params_def unit test that tests what the `get_conn_params` function. The function returns what postgres understands, i.e. user, password, port.
So i believe this test code is correct.

tests/integration/targets/setup_postgresql_db/tasks/main.yml
Andersson007: i've no idea why, but it removes only the first item with loop_control and our tests fail when trying to initialize DB and the directory is not empty. This happened locally, and this change helped.

tests/unit/plugins/module_utils/test_postgres.py
Andersson007: it looks confusing, yes, was to me too, we should probably rewrite the logic one day

andreasscherbaum (approved):

Andersson007:
Thanks for reviewing and merging folks!

PR #886, opened 2025-10-04
Feature inv/lookup

gebz97:
##### SUMMARY
1. Inventory plugin using PostgreSQL as the dynamic inventory source
2. Lookup plugin for pulling data from a PostgreSQL table

##### ISSUE TYPE
- New Plugin Pull Request

##### COMPONENT NAME
1. postgresql_inventory
2. postgresql_lookup

##### ADDITIONAL INFORMATION
The purpose of the plugins are pretty straight forward, I am unsure however if it is a must for me to include integration tests for the lookup plugin (I did include 26 unit tests for the inventory plugin).
(this is my first contribution to this project, so please be patient with me :) )


toydarian:
Full disclosure: I haven't looked at the code yet.

> I see issues that need to be solved, but let's not review the code and first discuss whether we need the plugins as it's big chunks of code that need to be maintained. I wouldn't expand the collection unless proposed features would be really helpful for a broad audience.

I can definitely see how this would be useful. But I don't know about the broader audience. Maybe @gebz97 can shine some light on that?  
Regarding the maintenance, I would definitely want to see integration tests, as that is an amount of code that for sure hides some bug somewhere. Integration tests are not necessarily there to show that the code doesn't contain issues at the moment, but more to ensure we don't break anything when we make changes later.  
I would also suspect that there are quite a few edge-cases that will come up if other people than the author use those plugins.

gspanos:
@andreasscherbaum 

> Having a database holding the inventory asks for trouble, as somehow this database must be installed and maintained.
> That's a job for Ansible. Which results in a chicken-egg problem.

Not sure exactly why is this a problem. Can't 2 playbooks have separate inventories? So you could have a static inventory for setting up the PostgreSQL and then, for the rest of the playbooks, rely on the PostgreSQL-based one?

gebz97:
Hello everyone,

Thanks for the feedback, I really appreciate it.

I will close this PR for now, but will definitely discuss all the points raised in the comments first on the matrix channel, and will proceed with creating separate PRs for each feature along with the justification/use cases.

If someone is interested in sharing insights on the design of the plugins I would appreciate it, as this is a first time for me.

PR #892, opened 2025-11-03
postgresql_privs: enhance type=table documentation

### LINKED ISSUE #841: Should be able to manage type 'view' in the privs module
##### SUMMARY

The addition of the type: 'view' would make it easier and also perhaps idempotent to grant access to a users instead of using the query module multiple times to first check the privileges and then add them, etc.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

postgresql_privs

##### ADDITIONAL INFORMATION

It would be super valuable to us if it was possible to manage privileges for the type 'view'. 

We have metric collectors that needs the access to be able to run SELECT queries against views, for example 'pg_stat_database' and 'pg_stat_bgwriter'.
At the moment we need to use a lot of other tasks using other modules to make the tasks idempotent. 
Therefore we would love to see that addition to the 'type' param.

For example a task could look like this:

```
- name: Grant SELECT privilege on pg_stat_database
  community.postgresql.postgresql_privs:
    login_db: postgres
    schema: public
    obj: pg_stat_database
    type: view
    roles: metrics_collector_user
    privs: SELECT
```

Instead of what we need to do now (to make it impodent)
````
- name: Check if SELECT privilege on pg_stat_database view already exists
  community.postgresql.postgresql_query:
    db: your_database
    login_user: your_superuser
    query: |
      SELECT 1
      FROM information_schema.role_table_grants
      WHERE grantee = 'your_user'
      AND table_name = 'pg_stat_database'
      AND privilege_type = 'SELECT'
  register: privilege_check
  become: true
  become_user: postgres

- name: Grant SELECT privilege on pg_stat_database view
  community.postgresql.postgresql_query:
    db: your_database
    login_user: your_superuser
    query: GRANT SELECT ON pg_catalog.pg_stat_database TO your_user
  when: privilege_check.rowcount == 0
  become: true
  become_user: postgres
```

<!--- HINT: You can also paste gist.github.com links for larger files -->

lalmeras:
##### SUMMARY
`type=table` in postgresql_privs can be used to manage view or materialized view privileges.
Make this behavior explicit in `postgresql_privs` in documentation.

Fixes #841

##### ISSUE TYPE
- Docs Pull Reques

##### COMPONENT NAME
community.postgresql / postgresql_privs

### DIFF
--- a/plugins/modules/postgresql_privs.py
+++ b/plugins/modules/postgresql_privs.py
@@ -48,7 +48,7 @@ options:
     - The C(type) choice is available since Ansible version 2.10.
     - The C(procedure) is supported since collection version 1.3.0 and PostgreSQL 11.
     - The C(parameter) is supported since collection version 3.1.0 and PostgreSQL 15.
-    - The C(table) is inclusive of foreign tables since collection version 3.6.0.
+    - The C(table) also includes views and materialized views. It is inclusive of foreign tables since collection version 3.6.0.
     type: str
     default: table
     choices: [ database, default_privs, foreign_data_wrapper, foreign_server, function,


Andersson007 (approved):

Andersson007:
@lalmeras thanks for the contribution!
@andreasscherbaum thanks for reviewing and approving it!

PR #930, opened 2026-06-01
Allow selective unquoted user configuration values

aleszeleny:
##### SUMMARY
Add `quote_configuration_values_exclude` to `postgresql_user` module so settings such as `search_path` can be passed unquoted while other configuration values keep the default quoting behavior. 
Preserve idempotency with enabled `reset_unspecified_configuration` by escaping quoted values only when building the `ALTER ROLE` query.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`postgresql_user`

##### ADDITIONAL INFORMATION
I've struggled with how to stick with `reset_unspecified_configuration` for ensuring inventory settings and follow the default recommended `quote_configuration_values` setting with `search_path` settings, as it must not be quoted to be correctly set.

An initial attempt was to split the configuration into two tasks, one for generic quoted settings and a special dedicated task for settings (so far I've faced only the `search_path`), but such a two steps approach doesn't work (as expected) with `reset_unspecified_configuration` option, which is what I want to use, earlier task settings were lost.

That brings me the idea of adding a parameter `quote_configuration_values_exclude` (or whatever more suitable name) that allows user configre an override for the `quote_configuration_values` for user provided list of parameters, that doesn't work (or use deceide on) with quoting so the module can still set all of them in one invocation, therefore making it "compatible" with `reset_unspecified_configuration` set to true.

Test case of the present behavior:

Create an Ansible superuser:
```sql
CREATE ROLE ansible_sup WITH password '1234' superuser login;
```
For the test on `localhost`, I've used just trust auth for local connection.

Validate connection:
```sh
psql -h /var/run/postgresql -d postgres -U ansible_sup -c "SELECT CURRENT_USER, version()"
```

Sample playbook:
```yml
---
- name: Configure PostgreSQL user search_path
  hosts: localhost
  gather_facts: false

  tasks:
    - name: Configure usr_t2 search_path
        login_host: /var/run/postgresql
        login_user: ansible_sup
        login_db: postgres
        name: usr_t1
        configuration:
          search_path: "nsp_1,nsp_2,public"
        quote_configuration_values: false
        reset_unspecified_configuration: true

    - name: Configure usr_t2 search_path
        login_host: /var/run/postgresql
        login_user: ansible_sup
        login_db: postgres
        name: usr_t2
        configuration:
          search_path: "nsp_1, nsp_2, public"
        quote_configuration_values: true
        reset_unspecified_configuration: true
```

Apply changes:
```bash
ansible-playbook -i lcoalhost, search_path_testcase.yml
```

Result:
```bash
psql -h /var/run/postgresql -d postgres -U ansible_sup -c "SELECT setrole::regrole, setconfig FROM pg_catalog.pg_db_role_setting WHERE setrole::regrole::text ~ 'usr_t?';"
Timing is on.
Null display is "«¤»".
Field separator is "|".
Output format is aligned.
Border style is 2.
Line style is unicode.
┌─────────┬──────────────────────────────────────────┐
│ setrole │                setconfig                 │
├─────────┼──────────────────────────────────────────┤
│ usr_t1  │ {"search_path=nsp_1, nsp_2, public"}     │
│ usr_t2  │ {"search_path=\"nsp_1, nsp_2, public\""} │
└─────────┴──────────────────────────────────────────┘
(2 rows)

Time: 1.025 ms
```

Hopefully, it'll be useful at least as an enhancement idea.

Kind regards Ales

### DIFF
--- a/changelogs/fragments/quote-configuration-values-exclude.yml
+++ b/changelogs/fragments/quote-configuration-values-exclude.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - postgresql_user - add the ``quote_configuration_values_exclude`` option to keep selected configuration values
+    unquoted while preserving default quoting for other values.
--- a/plugins/modules/postgresql_user.py
+++ b/plugins/modules/postgresql_user.py
@@ -145,9 +145,9 @@ options:
       - Removes any parameter from the user that is not listed here.
       - Parameters that are present in the database but are not included in this list will only be reset, if
         O(reset_unspecified_configuration=true).
-      - Inputs to O(user) as well as keys and values in this parameter are quoted by the module. If you require the
-        user to contain a C("), you need to double it, otherwise the module will fail. C(") and C(') are not allowed in
-        configuration keys and any C(') in the value of a configuration will be escaped by this module.
+      - Inputs to O(user) as well as keys and values in this parameter are quoted by the module by default. If you
+        require the user to contain a C("), you need to double it, otherwise the module will fail. C(") and C(') are
+        not allowed in configuration keys and any C(') in quoted configuration values will be escaped by this module.
         Additionally, parameters and values are checked if O(trust_input) is C(false).
     type: dict
     default: {}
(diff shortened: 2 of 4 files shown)


andreasscherbaum:
I'm unclear which problem this proposal solves. Can you please explain what the underlying issue is? Or even better: create a separate issue and reference the issue here.

aleszeleny:
> I'm unclear which problem this proposal solves. Can you please explain what the underlying issue is? Or even better, create a separate issue and reference it here.

Good point, I've tried to improve the description and testcase as well. Issue https://github.com/ansible-collections/community.postgresql/issues/931 was created.

Hope this is a more complete description.

PR #932, opened 2026-06-08
ci: add stable-2.21 to Azure Pipelines

preko-p:
## Summary
- add stable-2.21 sanity and unit jobs to Azure Pipelines
- add stable-2.21 Docker and remote jobs using the current devel target set
- list ansible-core 2.21 in the README tested versions

Fixes #927

## Testing
- `ruby -e "require 'yaml'; YAML.load_file('.azure-pipelines/azure-pipelines.yml')"`
- `git diff --check`
- Ruby structural check confirming `Ansible_2_21`, `Docker_2_21`, and `Remote_2_21` exist and are listed in `Summary.dependsOn`

## AI assistance disclosure
I used AI tools to help inspect the existing CI pattern and prepare this focused CI matrix update. I reviewed the resulting diff and ran the checks listed above.

PR #934, opened 2026-06-12
fix(CI): add retry with cache cleanup for RHEL langpack installation

Andersson007:
## Summary

Fixes failures that showed up in https://github.com/ansible-collections/community.postgresql/pull/932

  - Wrap the RHEL langpack install in `block/rescue` to handle transient
    mirror failures during RHEL point-release transitions (e.g., 10.1 → 10.2)
  - On failure, clean the DNF cache and retry

  RHEL repos start serving newer point-release packages before all CDN
  mirrors have the actual RPM files, causing `glibc-langpack-*` downloads
  to fail and breaking CI on RHEL 10.1 and 9.7 remote targets.

### DIFF
--- a/.azure-pipelines/azure-pipelines.yml
+++ b/.azure-pipelines/azure-pipelines.yml
@@ -183,6 +183,19 @@ stages:
             - name: RHEL 9.7
               test: rhel/9.7
 
+  - stage: Remote_2_21
+    displayName: Remote 2.21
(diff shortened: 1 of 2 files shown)


andreasscherbaum (approved):

Andersson007:
@andreasscherbaum thanks for reviewing and approving!

PR #937, opened 2026-06-26
feat(postgresql_membership): add admin_option, inherit_option and set_option

### LINKED ISSUE #757: postgresql_membership doesn't recognize ADMIN, SET and INHERIT options
##### SUMMARY

Make it possible to control the ADMIN, SET and INHERIT options when GRANTing roles.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

community.postgresql.postgresql_membership

##### ADDITIONAL INFORMATION

When a non-superuser creates a role in PostgreSQL 16, it is automatically granted the new role only with ADMIN option set. However, that's not enough to create objects owned by the user. To quote the [documentation](https://www.postgresql.org/docs/16/sql-grant.html):

> To create an object owned by another role or give ownership of an existing object to another role, you must have the ability to SET ROLE to that role; otherwise, commands such as ALTER ... OWNER TO or CREATE DATABASE ... OWNER will fail.

In earlier versions of PostgreSQL, I simply could use the `community.postgresql.postgresql_membership` to grant the new role to the non-superuser, but that doesn't work anymore because the module doesn't do anything now because the grant is there (albeit not with the options I need).

In my scenario, I cannot use a superuser to do this (superuser wouldn't receive this automatic grant) because I'm using a managed database as a service.

```yaml
- name: Demo
  hosts: integration
  gather_facts: true
  tasks:
    - name: Create role
        name: ansible_new_role
        password: dummy123!
        login_host: example.com
        login_db: postgres
        login_user: dbadmin
        login_password: password
        login_port: 5432
    - name: Grant new role to the admin user
      community.postgresql.postgresql_membership:
        group: ansible_new_role
        target_role: dbadmin
        login_host: example.com
        login_db: postgres
        login_user: dbadmin
        login_password: password
        login_port: 5432
    - name: Create database owned by the new role
      community.postgresql.postgresql_db:
        name: ansible_new_database
        owner: ansible_new_role
        login_host: example.com
        login_user: dbadmin
        login_password: password
        login_port: 5432
```

```
$ ansible-playbook play.yml -v -i inventory.yml 
Using /etc/ansible/ansible.cfg as config file

PLAY [Demo] **********************************************************************************************************************************************************************

TASK [Gathering Facts] ***********************************************************************************************************************************************************
[WARNING]: Platform linux on host integration is using the discovered Python interpreter at /usr/bin/python3.6, but future installation of another Python interpreter could
change the meaning of that path. See https://docs.ansible.com/ansible-core/2.16/reference_appendices/interpreter_discovery.html for more information.
ok: [integration]

TASK [Create role] ***************************************************************************************************************************************************************
changed: [integration] => changed=true 
  queries:
  - 'ALTER USER "ansible_new_role" WITH ENCRYPTED PASSWORD %(password)s '
  user: ansible_new_role

TASK [Grant new role to the admin user] ******************************************************************************************************************************************
ok: [integration] => changed=false 
  granted:
    ansible_new_role: []
  groups:
  - ansible_new_role
  queries: []
  state: present
  target_roles:
  - dbadmin

TASK [Create database owned by the new role] *************************************************************************************************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: psycopg2.errors.InsufficientPrivilege: must be able to SET ROLE "ansible_new_role"
fatal: [integration]: FAILED! => changed=false 
  msg: |-
    Database query failed: must be able to SET ROLE "ansible_new_role"

PLAY RECAP ***********************************************************************************************************************************************************************
integration                : ok=3    changed=1    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0   
```

It seems to me that this is partly a bug report because the module would otherwise run a regular `GRANT` statement which defaults to `SET` and `INHERIT` options enabled. In my case, however, it doesn't do anything because it sees the implicit `GRANT` with the `ADMIN` option only. There's the obvious workaround to use the `community.postgresql.postgresql_query` to apply the `GRANT` no matter what, but that's not clean.

Hoeze:
##### SUMMARY

Adds `admin_option`, `inherit_option` and `set_option` parameters to `postgresql_membership`, letting you control the `ADMIN`, `INHERIT` and `SET` options when granting roles.

Each option is a tri-state bool:
- `true` - ensure the option is enabled
- `false` - ensure the option is disabled
- unset (`null`, default) - use the PostgreSQL default for a new grant, keep the current value for an existing one

Fixes #757.

##### ISSUE TYPE

Feature Pull Request

##### COMPONENT NAME

postgresql_membership

##### ADDITIONAL INFORMATION

The feature commit is the original work of Felix Hamme (@betanummeric) from the `membership-options` branch referenced in #757; it is preserved as-is for attribution. The follow-up commits make it pass CI and fix two regressions found while preparing it:

- Added `version_added: '4.3.0'` to the three new options (required by `validate-modules`) and removed trailing whitespace from their descriptions.
- Restored plain `"%s"` quoting of role names in `GRANT`/`REVOKE`. The original commit switched to `pg_quote_identifier(name, 'role')`, which splits on dots and raises `SQLParseError` for role names containing them (e.g. `group.with.dots`), breaking the existing dotted-role integration tests. Role names are single identifiers; this matches how `postgresql_owner` quotes roles.
- Fixed `granted[]` over-reporting: `self.changed` was accumulated with `|=` and never reset, so once any grant fired every later role was reported as granted even when unchanged.
- Made the **option** path grantor-aware on PostgreSQL 16+, without changing plain membership behaviour:
  When an option is requested, the module compares against and updates the grant made by the *current role* (`__current_grant_options`), so an existing grant by another role does not suppress issuing the module's own grant with the requested options.
  Membership *existence* (`state` handling without options) is unchanged: `__fetch_members` still reports membership regardless of grantor, exactly as before, so there is no behaviour change for existing playbooks.
  Grantor-awareness is scoped to where it is actually needed (the options, which are PostgreSQL 16+ only); on PostgreSQL 16 a pair can be granted by several roles independently, so comparing the unfiltered membership's options would be order-dependent.
- Added a PostgreSQL >= 16 gated integration test covering:
  - grant-with-options
  - read-back from `pg_auth_members`
  - idempotency
  - single-option change
  - `granted[]` reporting fix
  - `state=exact` option reconciliation
  - grantor-isolation scenario from #757: a grant by another role does not suppress the module's own grant
  - PostgreSQL <16 test asserting the new options are rejected with an error

`validate-modules`, `pep8` and `pylint` sanity pass on the changed files. (The 4 failures in `tests/unit/.../test_postgres.py` in some local environments are pre-existing on `main` and unrelated to this change.)

###### Known limitations / possible follow-ups

All three options require PostgreSQL 16+, including `admin_option`, even though `WITH ADMIN OPTION` predates 16, for simplicity.

###### AI assistance disclosure

The version_added/whitespace/quoting/reporting/grantor fixes, the integration tests, and this description were produced with AI tooling assistance. The original feature commit is the work of @betanummeric. I have reviewed everything and take full responsibility for the contribution.

### DIFF
--- a/plugins/module_utils/postgres.py
+++ b/plugins/module_utils/postgres.py
@@ -384,7 +384,7 @@ class PgMembership(object):
         self.changed = False
         self.__check_roles_exist()
 
-    def __grant(self, group, role):
+    def __grant(self, group, role, role_obj=None):
         """Grant membership of group to role and apply the wanted options.
 
         With options (PostgreSQL 16+) only the grant made by the current role is
@@ -396,10 +396,12 @@ class PgMembership(object):
         Args:
             group (str) -- role whose membership is granted.
             role (str) -- role that receives the membership.
+            role_obj (PgRole) -- role's already-fetched membership state, to
+                avoid a redundant DB round trip when the caller already has it.
 
         Returns True when a change was made (bool).
         """
-        role_obj = PgRole(self.module, self.cursor, role)
+        role_obj = role_obj or PgRole(self.module, self.cursor, role)
         if group in role_obj.memberof:
             current = self.__current_grant_options(group, role) if self.options else {}
             if current is not None and all(current[option] == wanted
@@ -447,11 +449,12 @@ class PgMembership(object):
                     SET=rows[0]['set_option'])
 
     def grant(self):
+        role_objs = {role: PgRole(self.module, self.cursor, role) for role in self.target_roles}
         for group in self.groups:
             self.granted[group] = []
 
             for role in self.target_roles:
-                if self.__grant(group, role):
+                if self.__grant(group, role, role_objs[role]):
                     self.granted[group].append(role)
 
         return self.changed
@@ -495,7 +498,7 @@ class PgMembership(object):
             # options of memberships that already exist, so it runs for all
             # desired groups, not only the newly added ones.
             for group in desired_groups:
-                if self.__grant(group, role):
+                if self.__grant(group, role, role_obj):
                     if group in self.granted:
                         self.granted[group].append(role)
                     else:
--- a/plugins/module_utils/postgres.py
+++ b/plugins/module_utils/postgres.py
@@ -348,22 +348,22 @@ class PgRole():
         self.memberof = self.__fetch_members()
 
     def __fetch_members(self):
-        query = """select p.rolname, m.* -- include option columns if available (postgres >=16)
-                   from pg_catalog.pg_auth_members m
-                   join pg_catalog.pg_roles p on m.roleid=p.oid
-                   join pg_catalog.pg_roles c on m.member=c.oid
-                   where c.rolname = %s;"""
+        """Return the set of roles self.name is a member of.
+
+        Membership is reported regardless of which role granted it. Membership
+        options are read per-grantor by PgMembership when needed.
+
+        Returns a set of role names (str).
+        """
+        query = """SELECT p.rolname
+                   FROM pg_catalog.pg_auth_members m
+                   JOIN pg_catalog.pg_roles p ON m.roleid = p.oid
+                   JOIN pg_catalog.pg_roles c ON m.member = c.oid
+                   WHERE c.rolname = %s;"""
 
         rows = exec_sql(self, query, query_params=(self.name,),
                         add_to_executed=False)
-        result = dict()
-        for row in rows:
-            result[row['rolname']] = dict(
-                ADMIN=row.get('admin_option'),
-                INHERIT=row.get('inherit_option'),
-                SET=row.get('set_option'),
-            )
-        return result
+        return set(row['rolname'] for row in rows)
 
 
 class PgMembership(object):
@@ -373,8 +373,6 @@ class PgMembership(object):
         self.target_roles = [r.strip() for r in target_roles]
         self.groups = [r.strip() for r in groups]
         self.options = dict()
-        self.wanted_options = set()
-        self.unwanted_options = set()
         for option, setting in (options or dict()).items():
             if setting is not None:
                 self.options[option.upper()] = setting
@@ -387,16 +385,26 @@ class PgMembership(object):
         self.__check_roles_exist()
 
     def __grant(self, group, role):
+        """Grant membership of group to role and apply the wanted options.
+
+        With options (PostgreSQL 16+) only the grant made by the current role is
+        compared and updated, so a membership granted by another role (for
+        example the WITH ADMIN OPTION grant PostgreSQL creates when a
+        non-superuser creates a role) does not hide the need to issue our own
+        grant (see issue #757).
+
+        Args:
+            group (str) -- role whose membership is granted.
+            role (str) -- role that receives the membership.
+
+        Returns True when a change was made (bool).
+        """
         role_obj = PgRole(self.module, self.cursor, role)
-        # If role is in a group now, pass:
         if group in role_obj.memberof:
-            options_match = True
-            for option, desired_status in self.options.items():
-                if role_obj.memberof[group][option] != desired_status:
-                    options_match = False
-                    break
-            if options_match:
-                return False  # no change necessary
+            current = self.__current_grant_options(group, role) if self.options else {}
+            if current is not None and all(current[option] == wanted
+                                           for option, wanted in self.options.items()):
+                return False
 
         query = 'GRANT "%s" TO "%s"' % (group, role)
         changed = False
@@ -407,6 +415,36 @@ class PgMembership(object):
             changed |= exec_sql(self, q, return_bool=True)
         self.changed |= changed
         return changed
+
+    def __current_grant_options(self, group, role):
+        """Return the options of the group-to-role grant made by the current role.
+
+        Only reached on PostgreSQL 16+, where these columns exist and a pair can
+        be granted by several roles independently.
+
+        Args:
+            group (str) -- granted (group) role.
+            role (str) -- member role.
+
+        Returns a dict with ADMIN, INHERIT and SET keys, or None when the
+        current role has not granted this membership.
+        """
+        query = """SELECT m.admin_option, m.inherit_option, m.set_option
+                   FROM pg_catalog.pg_auth_members m
+                   JOIN pg_catalog.pg_roles g ON m.roleid = g.oid
+                   JOIN pg_catalog.pg_roles u ON m.member = u.oid
+                   WHERE g.rolname = %s
+                     AND u.rolname = %s
+                     AND m.grantor = (SELECT oid
+                                      FROM pg_catalog.pg_roles
+                                      WHERE rolname = CURRENT_ROLE);"""
+        rows = exec_sql(self, query, query_params=(group, role),
+                        add_to_executed=False)
+        if not rows:
+            return None
+        return dict(ADMIN=rows[0]['admin_option'],
+                    INHERIT=rows[0]['inherit_option'],
+                    SET=rows[0]['set_option'])
 
     def grant(self):
         for group in self.groups:
@@ -452,15 +490,16 @@ class PgMembership(object):
                 else:
                     self.revoked[group] = [role]
 
-            # 2. Filter out groups that in self.groups and
-            # the role is already member of and grant the rest
-            groups_to_grant = desired_groups - current_groups
-            for group in groups_to_grant:
-                self.__grant(group, role)
-                if group in self.granted:
-                    self.granted[group].append(role)
-                else:
-                    self.granted[group] = [role]
+            # 2. Ensure the role is a member of every desired group with the
+            # wanted options. __grant is idempotent and also reconciles the
+            # options of memberships that already exist, so it runs for all
+            # desired groups, not only the newly added ones.
+            for group in desired_groups:
+                if self.__grant(group, role):
+                    if group in self.granted:
+                        self.granted[group].append(role)
+                    else:
+                        self.granted[group] = [role]
 
         return self.changed
 
--- a/plugins/module_utils/postgres.py
+++ b/plugins/module_utils/postgres.py
@@ -21,8 +21,6 @@ from ansible.module_utils.common.text.converters import to_native
 from ansible.module_utils.basic import missing_required_lib
 from ansible_collections.community.postgresql.plugins.module_utils.version import \
     LooseVersion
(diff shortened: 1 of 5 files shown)


plugins/modules/postgresql_membership.py
```diff
@@ -46,6 +46,27 @@
       - If C(true), fail when group or target_role doesn't exist. If C(false), just warn and continue.
     default: true
     type: bool
+  admin_option:
+    description:
+      - This flag controls the membership option I(ADMIN). When unset (C(null)), the postgres default is used and existing configuration is preserved.
```
Andersson007:
```suggestion
      - This flag controls the membership option C(ADMIN). When unset (C(null)), the postgres default is used and existing configuration is preserved.
```
BTW, we should **in another pr** make the docs use Ansibel semantic markup. Just an idea for further contributions - many modules need that change.

plugins/modules/postgresql_membership.py
```diff
@@ -46,6 +46,27 @@
...
+      - Requires PostgreSQL 16 or later. Setting it against an older server makes the module fail. Ignored when I(state=absent).
+    default: null
+    type: bool
+    version_added: '4.3.0'
+  inherit_option:
+    description:
+      - This flag controls the membership option I(INHERIT). When unset (C(null)), the postgres default is used and existing configuration is preserved.
```
Andersson007:
```suggestion
      - This flag controls the membership option C(INHERIT). When unset (C(null)), the postgres default is used and existing configuration is preserved.
```

plugins/modules/postgresql_membership.py
```diff
@@ -46,6 +46,27 @@
...
+      - Requires PostgreSQL 16 or later. Setting it against an older server makes the module fail. Ignored when I(state=absent).
+    default: null
+    type: bool
+    version_added: '4.3.0'
+  set_option:
+    description:
+      - This flag controls the membership option I(SET). When unset (C(null)), the postgres default is used and existing configuration is preserved.
```
Andersson007:
```suggestion
      - This flag controls the membership option C(SET). When unset (C(null)), the postgres default is used and existing configuration is preserved.
```

tests/integration/targets/postgresql_membership/tasks/postgresql_membership_options.yml
```diff
@@ -0,0 +1,355 @@
...
+        SELECT m.set_option
+        FROM pg_catalog.pg_auth_members m
+        JOIN pg_catalog.pg_roles g ON m.roleid = g.oid
+        JOIN pg_catalog.pg_roles u ON m.member = u.oid
+        WHERE g.rolname = 'opt_group' AND u.rolname = 'opt_user1'
+    register: opts
+
+  - assert:
+      that:
+      - opts.query_result[0].set_option == false
```
Andersson007, line 327: In this and similar tasks where we change one option after setting a few, we should also check that the other options keep their previous values intact.

Andersson007 (commented):
@Hoeze thanks for the enhancement!
Could you please:
1. Add a couple of tasks with `check_mode: true` to check if the module correctly detects if any changes are needed / not needed and correctly reports it? I would also add one task to check with postgresql_query that no changes were actually made.
2. Add example for the new arguments to the EXAMPLES block
3. Take a look at my comment.

Andersson007:
@hunleyd thanks for merging some suggestion!

@Hoeze could you please take a look at https://github.com/ansible-collections/community.postgresql/pull/937#discussion_r3490662748 ?

Andersson007:
@Hoeze any updates on this?

Hoeze:
Dear @Andersson007, thank you for your feedback.
I was on vacation last week but I am having a look now :)

Hoeze:
Alright, I addressed your comments @Andersson007.
Please let me know what you think.

Andersson007 (approved):

Andersson007:
@Hoeze hi

> I was on vacation last week 

Cool, it was my assumption too, hope it was great and you've decided to become our regular contributor here:)

> Alright, I addressed your comments @Andersson007.
> Please let me know what you think.

LGTM now, thanks! Approved

@hunleyd @andreasscherbaum WDYT?

plugins/module_utils/postgres.py
```diff
@@ -377,20 +384,74 @@ def __init__(self, module, cursor, groups, target_roles, fail_on_role=True):
...
+        grant (see issue #757).
+
+        Args:
+            group (str) -- role whose membership is granted.
+            role (str) -- role that receives the membership.
+
+        Returns True when a change was made (bool).
+        """
+        role_obj = PgRole(self.module, self.cursor, role)
+        if group in role_obj.memberof:
```
hunleyd, line 414: Question on the membership model here: `__grant` (and `match`/`revoke` elsewhere in this class) uses `group in role_obj.memberof` as a grantor-blind check — it's true if the role is a member of the group at all, regardless of who granted it. But `__current_grant_options`, used a few lines below for comparing `admin_option`/`inherit_option`/`set_option`, is grantor-scoped — it only considers the grant made by the current connecting role (per the docstring, this is deliberate so a grant made by another role, e.g. an implicit WITH ADMIN OPTION grant, doesn't hide the need to issue our own grant per issue #757).

So the same method ends up applying two different definitions of "is this role a member of this group," depending on whether options are in play. Is that split intentional/necessary, or should the plain membership check also be scoped to the current role's grant for consistency? Just want to confirm this isn't going to bite someone who touches this code later without realizing there are two membership models tangled together here.

plugins/modules/postgresql_membership.py
```diff
@@ -79,6 +100,15 @@
     type: bool
     default: true
     version_added: '0.2.0'
+notes:
+- Membership existence (I(state) handling without the option parameters) considers
```
hunleyd: Related to my comment on `plugins/module_utils/postgres.py` around `__grant`: this note documents that passing `admin_option`/`inherit_option`/`set_option` silently switches the membership definition being checked, from grantor-blind to grantor-scoped. That means a previously idempotent task (no options set) can start returning `changed: true` and issuing a new GRANT the moment someone adds e.g. `set_option: true` — not because they asked for a different membership, but purely as a side effect of passing that param.

Is that switch-on-option-param behavior intentional as the long-term contract, or is it more of a necessary consequence of #757 that's worth a stronger callout (or a separate flag) so users don't get surprised by it? Wanted to flag it explicitly since it's easy to miss in the notes list.

plugins/module_utils/postgres.py
```diff
@@ -377,20 +384,74 @@ def __init__(self, module, cursor, groups, target_roles, fail_on_role=True):
...
+        non-superuser creates a role) does not hide the need to issue our own
+        grant (see issue #757).
+
+        Args:
+            group (str) -- role whose membership is granted.
+            role (str) -- role that receives the membership.
+
+        Returns True when a change was made (bool).
+        """
+        role_obj = PgRole(self.module, self.cursor, role)
```
hunleyd:
```suggestion
    def __grant(self, group, role, role_obj=None):
        """Grant membership of group to role and apply the wanted options.

        With options (PostgreSQL 16+) only the grant made by the current role is
        compared and updated, so a membership granted by another role (for
        example the WITH ADMIN OPTION grant PostgreSQL creates when a
        non-superuser creates a role) does not hide the need to issue our own
        grant (see issue #757).

        Args:
            group (str) -- role whose membership is granted.
            role (str) -- role that receives the membership.
            role_obj (PgRole) -- role's already-fetched membership state, to
                avoid a redundant DB round trip when the caller already has it.

        Returns True when a change was made (bool).
        """
        role_obj = role_obj or PgRole(self.module, self.cursor, role)
```
`__grant` currently re-fetches `PgRole` (a DB round trip) on every call. Accepting an already-fetched `role_obj` lets callers that already have it (see the next two suggestions) skip the redundant fetch.

plugins/module_utils/postgres.py
```diff
@@ -377,20 +384,74 @@ def __init__(self, module, cursor, groups, target_roles, fail_on_role=True):
...
-                role_obj = PgRole(self.module, self.cursor, role)
-                # If role is in a group now, pass:
-                if group in role_obj.memberof:
-                    continue
-
-                query = 'GRANT "%s" TO "%s"' % (group, role)
-                self.changed = exec_sql(self, query, return_bool=True)
-
-                if self.changed:
+                if self.__grant(group, role):
```
hunleyd:
```suggestion
    def grant(self):
        role_objs = {role: PgRole(self.module, self.cursor, role) for role in self.target_roles}
        for group in self.groups:
            self.granted[group] = []

            for role in self.target_roles:
                if self.__grant(group, role, role_objs[role]):
```
Each role's `PgRole` only needs fetching once, reused across every group in the outer loop, instead of once per `(group, role)` pair.

plugins/module_utils/postgres.py
```diff
@@ -429,16 +490,16 @@ def match(self):
...
-                if group in self.granted:
-                    self.granted[group].append(role)
-                else:
-                    self.granted[group] = [role]
+            # 2. Ensure the role is a member of every desired group with the
+            # wanted options. __grant is idempotent and also reconciles the
+            # options of memberships that already exist, so it runs for all
+            # desired groups, not only the newly added ones.
+            for group in desired_groups:
+                if self.__grant(group, role):
```
hunleyd:
```suggestion
                if self.__grant(group, role, role_obj):
```
`match()` already built `role_obj` for this role a few lines up; pass it through instead of letting `__grant` re-fetch it.

hunleyd (changes requested):
lmk if you have any questions on my comments

Andersson007:
@Hoeze when the PR is ready for review again, please state that explicitly. Thank you

Hoeze:
@Andersson007 yes, I'm not ready yet. I will request a new review once I am done:

From @hunleyd 's feedback, I think it would make sense to also implement the grantor explicitly. This is what i am preparing a stacked PR for. Please stay tuned :)

Andersson007:
> @Andersson007 yes, I'm not ready yet. I will request a new review once I am done:
> 
> From @hunleyd 's feedback, I think it would make sense to also implement the grantor explicitly. This is what i am preparing a stacked PR for. Please stay tuned :)

@Hoeze , cool, thanks, i think in this case it would make sense to click "Convert to draft" link below the CI runs box at the bottom of this conversation

Hoeze:
Superseded by #944 . There I implemented full support for the grantor model.
(In the end, I more-or-less started from scratch ^^)

PR #941, opened 2026-07-20
docs: update semantic markup in db, user, and query modules

### LINKED ISSUE #938: Check the DOCUMENTATION and RETURNS blocks in modules and make sure they use Ansible semantic markup
##### SUMMARY
Check the DOCUMENTATION and RETURNS blocks in modules (files under `plugins/modules`) and make sure they use Ansible semantic markup (some newer use it, but the majority doesn't).

https://docs.ansible.com/projects/ansible/latest/dev_guide/ansible_markup.html#semantic-markup-within-module-documentation

Atharva-Ramawat:
##### SUMMARY
Updated the `DOCUMENTATION` and `RETURNS` blocks to use proper Ansible semantic markup across three modules. Replaced outdated `I()` option tags with `O()`, updated boolean and specific string values to `V()`, and return values to `RV()`.

Fixes #938

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
`postgresql_db`, `postgresql_user`, `postgresql_query`

##### ADDITIONAL INFORMATION
Reviewed the modules to ensure compliance with the latest Ansible semantic markup guidelines (https://docs.ansible.com/projects/ansible/latest/dev_guide/ansible_markup.html). Carefully avoided nested markup to prevent `antsibull-docs` parser crashes, while ensuring literal syntax characters and SQL commands remained as `C()`.

### DIFF
--- a/plugins/modules/postgresql_db.py
+++ b/plugins/modules/postgresql_db.py
@@ -101,7 +101,7 @@ options:
   target:
     description:
     - File to back up or restore from.
-    - Used when O(state) is V(dump) or C(restore).
+    - Used when O(state) is V(dump) or V(restore).
     type: path
     default: ''
   target_opts:
--- a/plugins/modules/postgresql_db.py
+++ b/plugins/modules/postgresql_db.py
@@ -130,7 +130,7 @@ options:
     default: ''
   dump_extra_args:
     description:
-      - Provides additional arguments when O(state) is C(dump).
+      - Provides additional arguments when O(state) is V(dump).
       - Cannot be used with dump-file-format-related arguments like ``--format=d``.
     type: str
     version_added: '0.2.0'
--- a/plugins/modules/postgresql_db.py
+++ b/plugins/modules/postgresql_db.py
@@ -39,13 +39,13 @@ options:
   lc_collate:
     description:
       - Collation order (LC_COLLATE) to use in the database
-        must match collation order of template database unless C(template0) is used as template.
+        must match collation order of template database unless V(template0) is used as template.
     type: str
     default: ''
   lc_ctype:
     description:
       - Character classification (LC_CTYPE) to use in the database (e.g. lower, upper, ...).
-      - Must match LC_CTYPE of template database unless C(template0) is used as template.
+      - Must match LC_CTYPE of template database unless V(template0) is used as template.
     type: str
     default: ''
   icu_locale:
@@ -58,7 +58,7 @@ options:
   locale_provider:
     description:
       - Specifies the provider to use for the default collation in this database (LOCALE_PROVIDER).
-      - Possible values are icu (if the server was built with ICU support) or libc.
+      - Possible values are V(icu) (if the server was built with ICU support) or V(libc).
       - By default, the provider is the same as that of the template.
     type: str
     default: ''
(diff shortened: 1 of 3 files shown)


plugins/modules/postgresql_db.py
```diff
@@ -73,41 +73,41 @@
...
     description:
-    - Used to forcefully drop a database when the I(state) is C(absent), ignored otherwise.
+    - Used to forcefully drop a database when the O(state) is V(absent), ignored otherwise.
     type: bool
     default: False
   target:
     description:
     - File to back up or restore from.
-    - Used when I(state) is C(dump) or C(restore).
+    - Used when O(state) is V(dump) or C(restore).
```
hunleyd:
```suggestion
    - Used when O(state) is V(dump) or V(restore).
```
`restore` is a literal value of the `state` option (see `choices`), same as `dump` just converted on this line. The very next option, `target_opts`, has the identical sentence fully converted to `V(dump) or V(restore)`.

plugins/modules/postgresql_db.py
```diff
@@ -130,15 +130,15 @@
     default: ''
   dump_extra_args:
     description:
-      - Provides additional arguments when I(state) is C(dump).
+      - Provides additional arguments when O(state) is C(dump).
```
hunleyd:
```suggestion
      - Provides additional arguments when O(state) is V(dump).
```
`dump` is a literal value of `state` and is converted to `V(dump)` everywhere else this PR touches it (the `state` option itself, `target`, `target_opts`), but left as `C(dump)` here.

hunleyd (changes requested):
2 actionable issues found; suggested fixes below.

Both are incomplete C()->V() semantic-markup conversions, inconsistent with sibling lines this same PR fully converted.

plugins/modules/postgresql_db.py
```diff
@@ -175,9 +175,9 @@
     support: partial
     details:
       - The module is not idempotent when O(state=dump) — it always invokes C(pg_dump) and writes
-        the C(target) file, reporting changed=true on every invocation.
+        the C(target) file, reporting C(changed=true) on every invocation.
```
hunleyd: `C(target)` should be `O(target)` for consistency — the same option reference a few lines up ("rename the database O(name) to O(target)") was correctly converted to `O()`.

```suggestion
        the O(target) file, reporting C(changed=true) on every invocation.
```

plugins/modules/postgresql_db.py
```diff
@@ -175,9 +175,9 @@
...
       - The module is not idempotent when O(state=restore) — it always executes the contents of the
-        C(target) file against the database and reports changed=true.
+        C(target) file against the database and reports C(changed=true).
```
hunleyd: Same inconsistency — this `C(target)` refers to the same `target` option.

```suggestion
        O(target) file against the database and reports C(changed=true).
```

hunleyd (changes requested):
Nice, careful markup pass overall — correct calls on keeping `C(check_mode)`/`C(environment)` unconverted since those are global Ansible keywords, not module options, rather than blindly turning every `I()` into `O()`. Two consistency gaps found (inline suggestions below), plus one line the conversion missed entirely that's outside this diff so it can't be inline-commented:

`plugins/modules/postgresql_db.py` line 169, in `notes:`:
```
- State C(dump) and C(restore) don't require I(psycopg) since ansible version 2.8.
```
should become:
```
- State V(dump) and V(restore) don't require C(psycopg) since ansible version 2.8.
```
`state`'s `C(dump)`/`C(restore)` values were converted to `V()` everywhere else in this file, and `psycopg` is a library name rather than a module option, so it should use `C()` instead of the deprecated `I()`.

Andersson007:
@Atharva-Ramawat hello, any updates on this PR and specifically on @hunleyd 's suggestions?

Andersson007 (approved):
@hunleyd any other suggestions or we are good now? PTAL

plugins/modules/postgresql_db.py
hunleyd:
```suggestion
- State V(dump) and V(restore) don't require C(psycopg) since ansible version 2.8.
```

plugins/modules/postgresql_query.py
```diff
@@ -90,7 +90,7 @@
       - The module executes arbitrary user-supplied SQL commands and cannot determine whether a given query
         would modify state. For non-read-only queries, state change is set based on the PostgreSQL
         C(statusmessage) heuristic (e.g., C(INSERT)/C(UPDATE)/C(DELETE) with non-zero affected rows
```
hunleyd:
```suggestion
        RV(statusmessage) heuristic (e.g., C(INSERT)/C(UPDATE)/C(DELETE) with non-zero affected rows
```

plugins/modules/postgresql_query.py
hunleyd:
```suggestion
    - Set the client encoding for the current session (e.g. V(UTF-8)).
```

plugins/modules/postgresql_user.py
hunleyd:
```suggestion
      - Default of V(prefer) matches libpq default.
```

hunleyd (changes requested):

hunleyd (approved):

Andersson007:
@Atharva-Ramawat , @hunleyd thanks for you contributions!

PR #944, opened 2026-08-16
[5.0.0] postgresql_membership: manage membership options and the granting role

Hoeze:
# SUMMARY

Makes `postgresql_membership` work correctly on PostgreSQL 16 and later, and adds the membership options requested in #757.

PostgreSQL 16 reshaped role membership: a `(group, role)` pair can be granted independently by several roles, and each grant carries its own `ADMIN`, `INHERIT` and `SET` options. A membership is therefore identified by the group, the target role **and** the granting role.

The module did not account for this. It compared the grant it found against `CURRENT_ROLE`, but PostgreSQL records the bootstrap superuser as the granting role of a grant made by *any* superuser. The comparison therefore failed for every connection except the bootstrap superuser: `state=present` reported a change on every run, and `state=absent` issued no `REVOKE` at all.

This PR makes the module name the granting role explicitly, so it recognises its own grant on a later run.

**New options** (all PostgreSQL 16+): `admin_option`, `inherit_option`, `set_option`, `granted_by`.

**New return values**: `grants` (every grant of the requested pairs, with its granting role) and `effective_options` (the options the target role actually holds, since PostgreSQL applies them as the union of all grants).

Also adds `pg_quote_name` to `database` module_utils. `pg_quote_identifier` splits on dots, so it rejects a role whose name contains one; role names are never schema qualified, so a dot is part of the name.

# ISSUE TYPE

- Bugfix Pull Request
- Feature Pull Request

# COMPONENT NAME

postgresql_membership

# ADDITIONAL INFORMATION

This supersedes #937, which carried Felix Hamme's original implementation of these options. The option names and semantics are carried over and Felix is credited as co-author on the commit; the implementation is rewritten around the per-grantor model.

**Breaking changes**, all in the changelog fragment:

- On PostgreSQL 16+ the module names the granting role explicitly. A task where the connecting role holds `ADMIN OPTION` only indirectly now needs `granted_by`.
- `state=absent` and `state=exact` no longer report a change for a membership they cannot revoke.

The integration tests cover the grantor model, the membership options, `granted_by`, input handling and return-value shape, and include a `session_role` case that exercises the non-superuser path.

I created this PR using AI, but I hand-checked every change thoroughly.


plugins/module_utils/postgres.py
```diff
@@ -340,108 +342,397 @@ def get_conn_params(module, params_dict, warn_db_default=True):
...
+        # postgresql_membership rejects them before constructing this object.
+        self.wanted_options = dict((name, setting)
+                                   for name, setting in (membership_options or dict()).items()
+                                   if setting is not None)
+
+        self.grantor = self.__resolve_grantor(granted_by) if self.per_grantor_membership else None
+
         self.__check_roles_exist()
 
+    def __resolve_grantor(self, granted_by):
```
hunleyd: resolves ONE grantor for the whole PgMembership instance in __init__ (line 402), applied to every group in the task. PostgreSQL's real rule (select_best_admin in its own acl.c) picks the grantor per group — a connecting role can hold ADMIN OPTION on group_a directly and on group_b only through an inherited role, needing a different grantor for each.

Non-superuser deploy runs groups=[group_a, group_b], target_role=alice on PG16. Deploy holds ADMIN OPTION on group_a directly, on group_b only via team. Module names deploy for both → group_b fails permission denied to grant privileges as role "deploy", whole transaction rolls back, group_a grant lost too.

It's even worse for cloud users: RDS/Aurora/Cloud SQL master roles are never rolsuper, so this hits the default admin role on every managed-Postgres PG16+ instance the moment it inherits (rather than directly holds) ADMIN OPTION. granted_by can't fix it — it's a single string, groups is a list, so a mixed-source multi-group task has no escape but splitting into one task per group, undocumented anywhere.

Fix needs a design call: either derive grantor per-group via one query against pg_auth_members, or have a fail fast pre-flight naming granted_by as the fix instead of surfacing PostgreSQL's raw permission error.

plugins/module_utils/postgres.py
hunleyd, line 1: PgMembership is 439 of 921 lines in a file every module in the collection shares, with exactly one consumer. Same option-name triple (ADMIN/INHERIT/SET ↔ admin_option/etc) is hand-restated 3-4 times across 2 files with different casing/spelling each time (postgres.py:359,500,610,629,711,725,730 + membership.py:324,363-367). one shared constant would collapse all of it. Recommends splitting into plugins/module_utils/membership.p either here or as a follow-up

hunleyd (changes requested):
state=exact now returns an entry per group instead of only changed ones which breaks result.granted == {} assertions.
state=present now grants when only auto-created via WITH ADMIN OPTION which flips changed status for existing installs and can fire handlers. both should be breaking changes.

plugins/module_utils/postgres.py
Hoeze: > either derive grantor per-group via one query against pg_auth_members, 

The issue with this option is that there can be multiple possible grantors that could be chosen for a group.
1) Which one will be chosen?
2) What if the possible grantors change lateron, s.t. on first run, "alpha" is chosen as grantor, but on a later run, "beta" is chosen? This would effectively lead to duplicating the grant.

This is why I chose to require an explicit `granted_by` in this case.

> or have a fail fast pre-flight naming granted_by as the fix instead of surfacing PostgreSQL's raw permission error.

I will rather implement this option, yes.

Hoeze:
Thanks a lot for your review @hunleyd!

> resolves ONE grantor for the whole PgMembership instance in init (line 402), applied to every group in the task.

On that note, right now the grants are defined as the product of all options.
I think it would make sense to allow passing an explicit list of options, one grant per item.
This could look like this:
```yaml
community.postgresql.postgresql_membership:
  login_db: mydb
  state: present
  target_roles: [alice, bob]     # default for every row
  inherit_option: true           # default for every row
  memberships:
    - groups: ro_group
      granted_by: app_admin
      admin_option: false
    - groups: rw_group
      granted_by: data_admin
      set_option: false
```

This would solve the "one grantor for all" issue and make the `state: "explicit"` option way more useful. What do you think?

hunleyd:
> This would solve the "one grantor for all" issue and make the `state: "explicit"` option way more useful. What do you think?

1. Overlap with existing params — proposal keeps top-level groups/granted_by/admin_option/set_option as "defaults for every row" alongside the new memberships list. That's two ways to express the same grant now. Needs an explicit precedence/mutual-exclusion rule (e.g. memberships present → top-level per-group options ignored, only target_roles/inherit_option apply as defaults) or users will hit ambiguous-merge bugs.
2. This doesn't auto-resolve grantor from pg_auth_members it's the "explicit escape hatch" option, just cleaner than one-task-per-group. Fine as a v1 — just don't let the PR description imply it "fixes" the underlying detection gap, since a user who doesn't know their inherited-vs-direct ADMIN OPTION situation still has to hand-split memberships correctly or hit the same permission-denied wall, just per list-item instead of per-group globally.

Hoeze:
> That's two ways to express the same grant now. Needs an explicit precedence/mutual-exclusion rule

I see three ways to tackle this issue:

1. **Make `memberships` mutually exclusive** with `groups`, `granted_by`, `admin_option`, `inherit_option` and `set_option`. `target_roles` stays allowed at the top level, because the grantor and the options belong to the group rather than the target role (`select_best_admin` takes the current role and the group):

    ```yaml
    target_roles: [alice, bob]
    memberships:
      - groups: read_only
        granted_by: app_team
      - groups: read_write
        granted_by: dba_team
        admin_option: false
    ```

    We could then also deprecate `groups`, `granted_by` and the three option parameters in favour of `memberships`.

2. **Treat the top-level values as defaults**, as `rules_behavior: combine` does in `postgresql_pg_hba`:

    ```yaml
    granted_by: app_team
    admin_option: false
    memberships:
      - groups: read_only       # inherits both, silently
      - groups: read_write
        granted_by: dba_team    # overrides one, still inherits admin_option
    ```

    The first row takes a grantor it never named, which is one grantor for every group and the bug this PR fixes.

3. **Add a new module taking only `memberships`** and deprecate this one, leaving one form to document. It does not remove the current work, since the PG16 fixes have to land in `postgresql_membership` either way, and we would ship both modules for a full major version.

I lean towards 1). What do you think @hunleyd?

hunleyd:
> I lean towards 1). What do you think @hunleyd?

Concur, but one gap in option 1 as written: doesn't say what happens to target_roles per membership row — confirm each memberships item still applies to the full target_roles set, not per-row target overrides, or that needs stating explicitly too.

plugins/module_utils/postgres.py
Hoeze: Moved it to its own module

hunleyd:
lgtm

Hoeze:
Alright, I updated the documentation and added a `granted_by_any` option.
This is the intended new behavior:

> From PostgreSQL 16 on, a role membership is not one fact but one grant per
> granting role, each carrying its own ``ADMIN``, ``INHERIT`` and ``SET``
> options. The ``memberships`` option of
> :ansplugin:`community.postgresql.postgresql_membership#module` therefore has
> to decide, for every row, whose grant it manages before it can decide what to
> change. This page shows that decision as one tree. Everything below assumes
> PostgreSQL 16 or later; before 16 a pair has exactly one grant and no options,
> so every branch collapses into the same behaviour.
> 
> Step 0 resolves, per row, whose grant it manages::
> 
>     granted_by_any (row, else task level when no granted_by) ?
>     |-- true ---------> manages EVERY grant of its pairs          [G = any]
>     |-- granted_by: X -> manages the grant recorded under X       [G = X]
>     '-- neither ------> manages the grant under the derived role  [G = derived]
>                         (bootstrap superuser if the connecting role is a
>                          superuser, else CURRENT_ROLE)
> 
> Then the task builds one wanted grant per (row, target role, group) and reads
> the server's grants of every pair once, before anything is emitted. All
> decisions below compare against that snapshot, and roles not named in
> ``target_roles`` are never touched.
> 
> ``state=present``, per wanted grant::
> 
>     G = any:  does ANY grant of the pair exist?
>               |-- yes -> nothing
>               '-- no --> GRANT group TO role            (server picks grantor)
> 
>     G = X or derived:  does the grant under G exist?
>               |-- no --> GRANT ... [WITH <options>] GRANTED BY G       (1)(2)
>               |-- yes, but a named option key differs from it
>               |        -> same GRANT again, restating every named option
>               '-- yes, options match or none named -> nothing
>                         (a foreign grant never satisfies the request; one that
>                          keeps a cleared option alive only draws a warning)
> 
>     (1) before the first statement, and only for grants that will be emitted:
>         the connecting role must have the privileges of every named X, and
>         G must hold ADMIN OPTION on the group (waived for the bootstrap
>         superuser); the failure names the roles that do hold it
>     (2) all wanted GRANTs are decided first, so one refused grant stops the
>         task before anything runs
> 
> ``state=absent``, per wanted grant::
> 
>     G = any:  one REVOKE ... GRANTED BY <grantor> per grant found, after
>               checking the privileges over every one of those grantors
> 
>     G = X or derived:  does the grant under G exist?
>               |-- yes -> REVOKE ... GRANTED BY G   (privileges of X checked)
>               '-- no --> nothing
>               plus a warning for every OTHER grant that keeps the membership
>               alive, whether or not something was revoked
> 
> ``state=exact`` is a prune pass followed by the ``state=present`` logic::
> 
>     per TARGET ROLE, every group it is a member of that no row names for it:
>         task-level granted_by_any ?
>         |-- true --> revoke EVERY grant of it (as G = any, privileges checked)
>         '-- false -> revoke the grant under the DERIVED role only;
>                      a foreign grant survives with a warning pointing at a
>                      state=absent task with granted_by
>     then every wanted grant is granted or reconciled exactly as in present.
> 
> Two properties hold throughout: ``changed`` is true exactly when at least one
> ``GRANT`` or ``REVOKE`` was emitted, and the whole task is one transaction, so
> a failure at any branch leaves the server untouched.

@hunleyd @Andersson007 would you mind having another look?

hunleyd (approved):
looks like you addressed all my comments

Andersson007:
@Hoeze @hunleyd I now have a very important question: do those changes under breaking_changes in the changelog fragment really break things? (sorry if i miss anything from the conversation, the time is limited these days).
I'm asking because we release breaking change in major releases. Are we ready for a major release?

@mariolenz if we decide to proceed with a minor one, could you remind me what kind of entry type should we use, major_changes or breaking_changes so that there's no problem on the Ansible package side later?

Hoeze:
@Andersson007 I do not feel qualified to make that judgement:
- The changes listed as "breaking" definitely change some behavior, but maybe they only affect a small enough number of users to not be worried about?
- At the same time, why not draft a v5?

hunleyd:
i'd feel 'better' about cutting a new major. the breakage might be 'small' but it's non-zero

Andersson007:
@Hoeze @hunleyd I'd like to release, but I have no time this week (well, maybe Friday). There are however a [few other issues that have to be done until then](https://github.com/ansible-collections/community.postgresql/issues?q=is%3Aissue%20state%3Aopen%20milestone%3A5.0.0) (associated with 5.0.0 milestone). I can't however spend much time even on Friday.
Would you folks like to help to handle them all until Friday. They are quite straightforward (but should be done as separate PRs) and mostly about removing some already deprecated aliases/option.
What do you think?

hunleyd:
> @Hoeze @hunleyd I'd like to release, but I have no time this week (well, maybe Friday). There are however a [few other issues that have to be done until then](https://github.com/ansible-collections/community.postgresql/issues?q=sort%3Aupdated-desc+is%3Aissue+state%3Aopen+milestone%3A5.0.0+) (associated with 5.0.0 milestone). I can't however spend much time even on Friday. Would you folks like to help to handle them all until Friday. They are quite straightforward (but should be done as separate PRs) and mostly about removing some already deprecated aliases/option. What do you think?

I won't have bandwidth myself as I'm the on-call fool for $dayjob :(

I could help late _next_ week (Sep 17 on)

Andersson007:
@hunleyd thanks!
I've merged the PR. In case you have no time, we should move the issues to 6.0.0 milestone (that doesn't exist yet) and  remember that the next release here is 5.0.0:)

Hoeze:
Thanks a lot @hunleyd and @Andersson007 :tada:

PR #945, opened 2026-08-16
Quote database object names as identifiers everywhere

Hoeze:
> **Stacked on #944.** Until that merges, the diff here also shows #944's commit. Please review #944 first; this PR will be rebased onto `main` and taken out of draft once it lands. The seven commits listed under Scope below are the ones belonging to this PR.

# SUMMARY

Makes this collection treat a database object name as a literal name everywhere, so a name containing a double quote produces valid SQL instead of a syntax error or, in one case, executed SQL.

# ISSUE TYPE

- Bugfix Pull Request

# COMPONENT NAME

postgresql_user, postgresql_owner, postgresql_privs, postgresql_db, postgresql_schema, postgresql_sequence, postgresql_idx, postgresql_table, postgresql_tablespace, postgresql_publication, postgresql_subscription, postgresql_info, postgres and database module_utils

# ADDITIONAL INFORMATION

**One breaking change.** `postgresql_user` read a name that both started and ended with a double quote as a pre-quoted identifier, and rejected a lone double quote with `User escaped identifiers must escape extra quotes`. The name is now taken literally, the way every other module in this collection treats a role name, so `"foo"` creates a role called `"foo"` rather than `foo`. Documented on the `name` option and in the changelog.

## Scope

Seven commits:

1. `fix(postgresql_owner,postgresql_privs): quote role names as identifiers`
2. `fix(postgresql_user): take the role name literally`
3. `fix(module_utils): quote the object name in COMMENT ON`
4. `fix(module_utils,postgresql_user,postgresql_db): escape names in parameterised statements`
5. `fix: quote identifiers in the remaining modules`
6. `test: cover names containing a double quote in the swept modules`
7. `test: run the quoted-name tests against the maintenance database`

Commits 1 to 4 fix `postgresql_user`, `_owner`, `_privs` and `_db`. Commit 5 sweeps the remaining modules, so the collection ends up consistent rather than half converged.

**For a name needing no escaping the generated SQL is byte-identical to before.**

## Testing

Commits 6 and 7 add integration coverage that creates objects whose names contain a double quote and reads them back from the catalog under the literal name, in `postgresql_schema`, `_sequence`, `_idx`, `_table` and `_db`, plus a `session_role` case for the shared `SET ROLE`. The existing `trust_input` tests only asserted that such a name is _rejected_, so none of them reached the quoting code.

Run against real PostgreSQL, both branches of every version gate:

| server | result |
| ------ | ------ |
| PostgreSQL 16.14 | all targets `failed=0` |
| PostgreSQL 13.23 | all targets `failed=0` |

Not covered: `postgresql_tablespace` (its target sets up no filesystem directory), `postgresql_publication` and `_subscription` (replication setup), and `postgresql_info` (the swept site is `SHOW` on a GUC name, which cannot contain a quote, so the quoting there is defensive).

## COMMENT ON, and names containing a percent sign

Most of the sites above produce a syntax error when a name contains a double quote. Two of them are worse than that, and are the reason for commits 3 and 4.

**`COMMENT ON` interpolated the object name.** `set_comment` parameterised the comment value but built the statement around a raw name. Under psycopg2, which substitutes client side, a name shaped to close the statement runs what follows:

```
name: 'postgres" IS $$x$$; CREATE TABLE pwned(i int); --'
-> COMMENT ON ROLE "postgres" IS $$x$$; CREATE TABLE pwned(i int); --" IS 'hi'
```

Under psycopg3 the same input is a hard error instead, since it binds server side. Reachable in `postgresql_user`, `_db`, `_schema`, `_ext`, `_publication`, `_subscription` and `_tablespace`. Fixed in commit 3.

**A percent sign in a name broke the statements that also carry parameters.** psycopg substitutes over the whole statement, so a percent sign left in an interpolated identifier opens a placeholder. `CREATE USER`, `ALTER USER`, `CREATE DATABASE` and `COMMENT ON` are all affected. Usually this just fails loudly with `unsupported format character`, which is how you would notice a name like `report_100%`. The bad case is a name shaped like a named placeholder:

```
name: 'x%(password)sy'  ->  role literally named  x'<the password>'y
```

which writes the password into `pg_authid` and into `executed_queries`. That shape is contrived rather than something you would hit by accident, but it is worth closing. `pg_quote_name` grows a `for_params` flag that doubles the percent sign, which psycopg then reduces back to one. Fixed in commit 4.

**`trust_input` does not reliably block either of those two.** It is a blocklist of three regexes, and `PATTERN_3` lists `SELECT|UPDATE|INSERT|DELETE|DROP|TRUNCATE|ALTER` but not `CREATE`, `GRANT`, `REVOKE` or `COMMENT`. The `COMMENT ON` payload above is caught only because it ends in `--`; drop that and it passes, as does a `GRANT` variant and the percent-sign case:

| input | `trust_input: false` blocks it |
| ----- | ------------------------------ |
| `postgres" IS $$x$$; CREATE TABLE pwned(i int); --` | yes |
| the same without the trailing `--` | no |
| `postgres" IS $$x$$; GRANT ... TO evil` | no |
| `x%(password)sy` | no |

That blocklist gap is a separate matter from this PR and affects all 22 modules exposing `trust_input`, so this PR does not touch it. It is the reason the quoting is worth fixing on its own terms rather than leaning on that option.

## AI notice
Note that this PR was created using AI


Andersson007:
@Hoeze hi, thank for doing this! Is it ready for review and the draft status is still relevant?

Hoeze:
Hi @Andersson007, this PR is basically ready but it depends on #944. I would wait until the other one got merged, before I ask for a review :)

Andersson007:
> Hi @Andersson007, this PR is basically ready but it depends on #944. I would wait until the other one got merged, before I ask for a review :)

@Hoeze great, thanks for clarifying! I see @hunleyd left another comment in https://github.com/ansible-collections/community.postgresql/pull/944#issuecomment-5395960429 a few days ago

Andersson007:
@Hoeze I've merged https://github.com/ansible-collections/community.postgresql/pull/944 , so you can proceed, thanks!

Andersson007:
@Hoeze hello, any updates on this? I've merged your previous PR (if it matters for finishing this one) and have just released the collection. Please let us know if you need any guidance.

Hoeze:
Hi @Andersson007, thanks for the ping!
I rebased this PR and it is ready for review now :)

plugins/modules/postgresql_privs.py
hunleyd:
```suggestion
        quoted_schema_qualifier = pg_quote_name(schema_qualifier) if schema_qualifier else None
```

hunleyd (changes requested):

plugins/modules/postgresql_privs.py
Hoeze: Makes sense!

hunleyd (approved):

andreasscherbaum (approved):

Andersson007:
Thanks for your contributions, everyone!

Andersson007:
@Hoeze now I see there's another `breaking_change`
Is it really breaking or not so much? Otherwise, I think now we should release 6.0.0

Hoeze:
Thank you @Andersson007 :raised_hands: 

> @Hoeze now I see there's another `breaking_change` Is it really breaking or not so much? Otherwise, I think now we should release 6.0.0

I just tried using the current v5.0.0 with manually-quoted names and it turns out that "hacking" the role this way does not work. These names just fail the role.
So no, it is actually no breaking change but rather a new feature. Maybe we should change that one :sweat_smile:

Andersson007:
@Hoeze great! could you please change breaking_changes to minor_changes in another PR then?

Hoeze:
@Andersson007 done: https://github.com/ansible-collections/community.postgresql/pull/958

PR #952, opened 2026-09-09
[5.0.0] remove!: remove deprecated db and database aliases of login_db from 19 modules

Andersson007:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.postgresql/issues/831

Assisted by AI

PR #954, opened 2026-09-11
ci: split Docker integration tests into two groups to prevent timeouts

Andersson007:
## Summary
- With all 21 test targets in `shippable/posix/group1`, Ubuntu Docker CI jobs exceed the 50-minute time limit when many targets need to run (e.g., PRs touching multiple modules like #952)
- Split into group1 (10 targets) and group2 (11 targets) so each job stays under the limit
- Updated Azure Pipelines config to run both groups for each Docker target

## Test plan
- [ ] Verify CI pipeline generates jobs for both group1 and group2 per Docker target
- [ ] Verify Ubuntu Docker jobs complete within the 50-minute limit
- [ ] Verify Fedora and RHEL jobs are unaffected

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.azure-pipelines/azure-pipelines.yml
+++ b/.azure-pipelines/azure-pipelines.yml
@@ -129,7 +129,10 @@ stages:
     jobs:
       - template: templates/matrix.yml
         parameters:
-          testFormat: devel/linux/{0}/1
+          testFormat: devel/linux/{0}
+          groups:
+            - 1
+            - 2
           targets:
             - name: Fedora 44
               test: fedora44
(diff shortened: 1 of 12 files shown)


Andersson007:
hm, in most Ubuntu jobs, it gets stuck on:
```
00:31 TASK [setup_postgresql_db : Run autoremove] ************************************
```
or
```
05:08 TASK [setup_postgresql_db : Install required OS packages] **********************
```
UPDATE: it's infra issue with Ubuntu on AZP

Andersson007:
It helped: almost all the Ubuntu jobs succeeded, but I don't like that there are twice more jobs now (even if the total number of test targets stay the same). Let's close this one for now and maybe re-open and merge it later if we run into same issue.

PR #958, opened 2026-09-21
docs: list the postgresql_user role-name change as a minor change

Hoeze:
- Moves the postgresql_user role-name entry in `changelogs/fragments/converge-role-quoting.yml` from `breaking_changes` to `minor_changes`. The entry text itself is unchanged.
- Follow-up to #945: I tested 5.0.0 and found that role names wrapped in double quotes by hand already failed, so no working playbook breaks from this change. See the [maintainer's request](https://github.com/ansible-collections/community.postgresql/pull/945#issuecomment-5760770835).


Andersson007 (approved):

Andersson007:
@Hoeze thank you!

